PR #9, opened 2015-01-17
API change, typo fixes.

Earnestly:
This just fixes the spelling of priority before it gets baked into the API and becomes unchangeable.


Cloudef:
Already fixed when you informed on irc :)

PR #20, opened 2015-09-21
path configuration - 2nd part, prevent clashing of ncurses symbol, xinerama check 

sleep-walker:


sleep-walker:
I removed that ncurses madness, other commits makes IMHO sense

Cloudef:
Looks good. Thanks.

kyrias:
`CMAKE_INSTALL_LIBDIR` can, and often is, relative, which will break this implementation. This should probably use `CMAKE_INSTALL_FULL_LIBDIR` instead.

sleep-walker:
CMAKE_INSTALL_FULL_LIBDIR works fine for me as well.

Cloudef:
Good point, the #define should point to CMAKE_INSTALL_FULL_LIBDIR, as it does in orbment as well for plugins.

Cloudef:
c4a55379 Thanks @kyrias

PR #24, opened 2016-03-18
Make protocol specifications follow the DTD

mikkeloscar:
Gets rid of these warnings when building:

    WARNING: XML failed validation against built-in DTD


Cloudef:
Actually, proper fix is to make it take wayland-protocols submodule / and allow building against systemwide install (like wlc allows) or even temporary fix right now would be to update protocols from there manually :)

Cloudef:
Thanks

PR #29, opened 2016-12-03
Add C-m -> Return bind

upwhere:
C-m is often bound to return and it's jarring to have all the
other keybinds but not this one


lib/renderers/curses/curses.c
```diff
@@ -335,6 +335,10 @@ poll_key(const struct bm_menu *menu, uint32_t *unicode)
             terminate();
             return BM_KEY_ESCAPE;
 
+        case 13: /* C-m */
```
Cloudef: Move this under case 10: /* return */ as the logic inside case is same

upwhere: Done & squashed

PR #32, opened 2018-04-16
Wayland renderer: implement bottom option

martinetd:
I tried to mimic what x11 does; this isn't super efficient because `menu->bottom` isn't actually set when we create the windows but I guess it works anyway (might flicker a bit? I can't see it if so)

That `wlr_layer_surface_v1_set_anchor` line is getting super long too, happy to do whatever you think is good with it.


Cloudef:
Look ok, thanks!

PR #33, opened 2018-04-27
More wayland fixes

martinetd:
- I had positioning a bit wrong with my previous bottom PR; I thought it was just positioned above the bar by the compositor but it was just rendering to the top of a hard-coded 32 pixels buffer and leaving the bottom blank.
The first commit fixes that by setting size when it changes and requests to not honor exclude zone so we always have the X11 behavior of being at the very bottom.
We technically should call `wl_surface_commit` after `set_size` but it will be committed alongside the buffer right afterwards so I think that's fine? @SirCmpwn might know better.

- grab_keyboard is needed when running bemenu interactively, we previously grabbed keyboard on window creation before input was taken and there is no easy way out of this grab...

- last commit is a nitpick, but might as well remove what's not needed.

Happy to split in two PR if you think it's worth it, or whatever.

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -82,7 +82,7 @@ usage(FILE *out, const char *name)
           "   c = ncurses, w == wayland, x == x11\n"
           "   (...) At end of help indicates the backend support for option.\n\n"
 
-          " -b, --bottom          appears at the bottom of the screen. (x)\n"
+          " -b, --bottom          appears at the bottom of the screen. (wx)\n"
           " -f, --grab            show the menu before reading stdin. (wx)\n"
           " -m, --monitor         index of monitor where menu will appear. (x)\n"
           " --fn                  defines the font to be used ('name [size]'). (wx)\n"
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -40,7 +40,7 @@ render(const struct bm_menu *menu)
     if (wayland->input.code != wayland->input.last_code) {
         struct window *window;
             wl_list_for_each(window, &wayland->windows, link) {
-            bm_wl_window_render(window, menu);
+            bm_wl_window_render(window, wayland->display, menu);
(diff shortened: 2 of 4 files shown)


SirCmpwn:
>We technically should call wl_surface_commit after set_size

Yeah do this, and probably roundtrip too

SirCmpwn:
Also while you're at it, you can remove the wl_output logic and update the protocol because passing NULL for an output now means "let the compositor decide".

martinetd:
>> We technically should call wl_surface_commit after set_size
>
> Yeah do this, and probably roundtrip too

Ok, that doesn't cost much to do, might as well play safe. Done.

> Also while you're at it, you can remove the wl_output logic and update the protocol because passing NULL for an output now means "let the compositor decide".

Hmm, that's fine (and removes a lot of code), but we lose track of max_height that was per output; if there's a single window = single surface/rendering pass and we don't know where it will be displayed we either have to pick the smallest screen's height there is... Or as a trick, calling `set_size` with 0 height apparently gives the chosen output height so we can use the first configure event to set that height... But that feels wrong to rely on that behaviour?
Not having a max height breaks the `-l` option.

(I also don't remember what behaviour we have on X11, does it display on active monitor by default? Will probably want to implement `-m` if we don't display on all outputs.. Will check tomorrow when I have another screen)

SirCmpwn:
>Or as a trick, calling set_size with 0 height apparently gives the chosen output height so we can use the first configure event to set that height... But that feels wrong to rely on that behaviour?

Nah this is an acceptable approach

Cloudef:
>(I also don't remember what behaviour we have on X11, does it display on active monitor by default? Will probably want to implement -m if we don't display on all outputs.. Will check tomorrow when I have another screen)
Active monitor, -m can is for choosing monitor

Cloudef:
Don't have the env for testing this right now. But the code looks good.

PR #36, opened 2018-11-07
Fix wayland renderer lagging behind by 1 keystroke

### LINKED ISSUE #35: Display of selected item lags behind
I use bemenu (bemenu-git from AUR) with sway using the wayland backend and open it using mod+d. When trying to launch for example "playonlinux" the sequence is as follows:

Key Pressed - Displayed Value - Displayed list
p - (NULL) - (everything)
pl - p - p11-kt, p11tool....
pla - pl - pl2pm, play, playonlinux
play - pla - play, playonlinux
playo - play - play, playonlinux

Now when I hit enter "playonlinux" is run although "play" is marked as the active entry. So it seems only the display lags behind by one while intenally everything is correct.

ppascher:
When using the wayland renderer the entries according to the second to last keystroke where displayed while the actual selected entry matched the last keystroke.
This removes skipping the buffer change in bm_wl_window_render if window->frame_cb is not NULL.
Fixes #35

### DIFF
--- a/lib/renderers/wayland/window.c
+++ b/lib/renderers/wayland/window.c
@@ -208,9 +208,6 @@ bm_wl_window_render(struct window *window, struct wl_display *display, const str
 {
     assert(window && menu);
 
-    if (window->frame_cb)
-        return;
-
     struct buffer *buffer;
     for (int tries = 0; tries < 2; ++tries) {
         if (!(buffer = next_buffer(window))) {


ppascher:
Not sure if this is the correct way to fix the issue but it works for me. Though I am sure there was a reason for implementing the window->frame_cb check...

Cloudef:
Did you check that this issue exists in weston as well?

ppascher:
I could not get bemenu to run in weston unfortunatly. Only with the curses backend.

Cloudef:
The frame_cb check is there to make sure we don't render another frame until the compositor has handled our last frame. (which shouldn't happen, since we wait for each frame to get drawn) While this patch has no problem, (the bemenu will abort if it runs out of available buffers), it might be working around a bug in compositor.

Cloudef:
For running in weston, you could rollback before this commit https://github.com/Cloudef/bemenu/commit/17aa5e21bfdf08adeb65e4f40cd3e059cf7432ba it should run in weston then again, albeit somewhere in middle of the screen :D

ppascher:
Thanks for the pointers. I rolled back to the parent of the commit you mentioned and the issue this patch addresses does not exist running under weston.

Cloudef:
Does the issue exist with that commit under sway? (assuming it runs there)

Cloudef:
Ah damn. Well chipping in @SirCmpwn and his thoughts. It may be worth checking out if the weston's simple examples work properly in sway too, IIRC the window.c is pretty much based on them. I haven't been doing much anything with wayland recently, so if it feels like this PR should be merged, I don't have anything against it either. Looking at the patch, it's not wrong, but it may be working around bug, that's all. Though it may also be that bemenu is relying on weston implementation detail, in which case the patch is more than welcome.

HoshiYamazaki:
The bug still appears, tested on https://aur.archlinux.org/packages/bemenu-git/ (Arch Linux + Sway)

Cloudef:
Maybe we should merge this, if the key persons aren't here.

PR #37, opened 2019-01-28
Fix vertical alignment of text.

v44r:
The text inside the box is below center, which causes some descents to be cut out. This centers the text.

### DIFF
--- a/lib/renderers/cairo.h
+++ b/lib/renderers/cairo.h
@@ -22,6 +22,7 @@ struct cairo_paint {
     struct cairo_color fg;
     struct cairo_color bg;
     const char *font;
+    int32_t baseline;
 
     struct box {
         int32_t lx, rx; // left/right offset (pos.x - lx, box.w + rx)
@@ -37,6 +38,7 @@ struct cairo_paint {
 struct cairo_result {
     uint32_t x_advance;
     uint32_t height;
+    uint32_t baseline;
 };
 
 struct cairo_paint_result {
@@ -104,10 +106,12 @@ bm_pango_get_text_extents(struct cairo *cairo, struct cairo_paint *paint, struct
     PangoRectangle rect;
     PangoLayout *layout = bm_pango_get_layout(cairo, paint, buffer);
     pango_layout_get_pixel_extents(layout, NULL, &rect);
+    int baseline = pango_layout_get_baseline(layout) / PANGO_SCALE;
     g_object_unref(layout);
 
     result->x_advance = rect.x + rect.width;
     result->height = rect.height;
+    result->baseline = baseline;
     return true;
 }
 
@@ -131,6 +135,7 @@ bm_cairo_draw_line(struct cairo *cairo, struct cairo_paint *paint, struct cairo_
     int width, height;
     pango_layout_get_pixel_size(layout, &width, &height);
     height = paint->box.h > 0 ? paint->box.h : height;
+    int base = pango_layout_get_baseline(layout) / PANGO_SCALE;
 
     cairo_set_source_rgba(cairo->cr, paint->bg.r, paint->bg.b, paint->bg.g, paint->bg.a);
     cairo_rectangle(cairo->cr,
@@ -140,7 +145,7 @@ bm_cairo_draw_line(struct cairo *cairo, struct cairo_paint *paint, struct cairo_
     cairo_fill(cairo->cr);
 
     cairo_set_source_rgba(cairo->cr, paint->fg.r, paint->fg.b, paint->fg.g, paint->fg.a);
-    cairo_move_to(cairo->cr, paint->box.lx + paint->pos.x, paint->pos.y);
+    cairo_move_to(cairo->cr, paint->box.lx + paint->pos.x, paint->pos.y - base + paint->baseline);
     pango_cairo_show_layout(cairo->cr, layout);
 
     g_object_unref(layout);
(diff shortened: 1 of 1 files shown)


Earnestly:
Yeah that was most likely introduced by https://github.com/Cloudef/bemenu/commit/b84ce02c7ecfb051943bc6a9ffd1d22d255995fb

The problem comes when trying to display pretty large glyphs and ensuring we get something reasonable.

Here is an example of how bad it can get: https://ptpb.pw/eRSR.png

A good test would be "༼ つ ◕_◕ ༽つ" or many of the examples on: https://textfac.es/

I've been using locally a patch which more or less works, but it's still a bodge:

```diff
diff --git a/lib/renderers/cairo.h b/lib/renderers/cairo.h
index 307b9c2..43d2f6c 100644
--- a/lib/renderers/cairo.h
+++ b/lib/renderers/cairo.h
@@ -140,7 +140,7 @@ bm_cairo_draw_line(struct cairo *cairo, struct cairo_paint *paint, struct cairo_
     cairo_fill(cairo->cr);
 
     cairo_set_source_rgba(cairo->cr, paint->fg.r, paint->fg.b, paint->fg.g, paint->fg.a);
-    cairo_move_to(cairo->cr, paint->box.lx + paint->pos.x, paint->pos.y + paint->box.ty);
+    cairo_move_to(cairo->cr, paint->box.lx + paint->pos.x, paint->pos.y + paint->box.ty - 1);
     pango_cairo_show_layout(cairo->cr, layout);
 
     g_object_unref(layout);
```

v44r:
The last commit removes the extra black bar below the last item in vertical mode. I think now all positioning/sizing makes sense and the text is nicely centered. 

I will make another patch to add configurable padding (instead of the fixed "2").

Earnestly:
With the new patch it looks fine except that when `-l` argument is larger than the available screen size I get these results:

l70

Honestly, I don't mind this as a compromise.  The author's test list also appears to display without issue regardless of the value I give `-l` and with both large and small fonts from `xos4 terminus 8` to `Inconsolatazi4 12`.

*Their test list includes almost 50,000 entries of such as the following sort of thing:*
```
六弦アリス >> 不思議の国の音哲樂　まやかし篇 >> 最果ての馬は霧の中を行く
六弦アリス >> 不思議の国の音哲樂　まやかし篇 >> 夢のつづき
大江戸宅急便 >> 月刊ミゾノクチ 5月号 >> 幽霊楽団　～DTMはじめました～
大江戸宅急便 >> 月刊ミゾノクチ 5月号 >> 恋風　～Long long time ago････～
中雑魚酒菜 >> 東方縦笛妖々夢 >> 独りぼっちの人形作り
中雑魚酒菜 >> 東方縦笛妖々夢 >> ネクロファンタジア
Rib:y >> 眞紅ノ眷屬ハ圓環スル螺旋ヲ嘆ズル >> 不死ナル王女ハ眞紅ノ冥淵ニテ哂ウ
Rib:y >> 眞紅ノ眷屬ハ圓環スル螺旋ヲ嘆ズル >> 諧謔ノ闇ハ終焉ヲ嘲ル凶音
```

Cloudef:
Not really happy with any of the results. The last image with -l over screen size is actually the best looking result. 

Ideally we'd just force a box size where we draw the text and let those tall characters clip, rather than messing the layout. However, I can see problem here if both tall and non-tall characters are mixed, as you can't find the "ideal base" from the text. It might be worth looking how terminals draw text per line.

v44r:
vte lets tall characters overlap:

image

They seem to use the set of all ASCII chars to calculate the height (VTE_DRAW_SINGLE_WIDE_CHARACTERS is an array with all the printable ASCII characters):

        /* Estimate for ASCII characters. */
        pango_layout_set_text (info->layout, VTE_DRAW_SINGLE_WIDE_CHARACTERS, -1);
        pango_layout_get_extents (info->layout, NULL, &logical);
        info->height = PANGO_PIXELS_CEIL (logical.height);

Not very elegant, but I don't see any other solution for a fixed height (cairo has cairo_font_extents() but I don't see anything similar for pango).

Cloudef:
That probably would work fine if you clip the drawing. Though it would be good idea to make sure CJK at least fits the box.

v44r:
Didn't see the last comment before committing. The last commit uses the same set of characters as vte to determine height. The lines are clipped.

Cloudef:
You could try screenshotting the list above posted by Earnestly.

```
六弦アリス >> 不思議の国の音哲樂　まやかし篇 >> 最果ての馬は霧の中を行く
六弦アリス >> 不思議の国の音哲樂　まやかし篇 >> 夢のつづき
大江戸宅急便 >> 月刊ミゾノクチ 5月号 >> 幽霊楽団　～DTMはじめました～
大江戸宅急便 >> 月刊ミゾノクチ 5月号 >> 恋風　～Long long time ago････～
中雑魚酒菜 >> 東方縦笛妖々夢 >> 独りぼっちの人形作り
中雑魚酒菜 >> 東方縦笛妖々夢 >> ネクロファンタジア
Rib:y >> 眞紅ノ眷屬ハ圓環スル螺旋ヲ嘆ズル >> 不死ナル王女ハ眞紅ノ冥淵ニテ哂ウ
Rib:y >> 眞紅ノ眷屬ハ圓環スル螺旋ヲ嘆ズル >> 諧謔ノ闇ハ終焉ヲ嘲ル凶音
```

In addition to the other kaomoji examples.
If these look fine, then I consider the PR good.

v44r:
The problem with mixing CJK with ASCII is that CJK moves the baseline a bit and ASCII descenders get cut. I'll try to see if I can draw text with the baseline in the correct position independently of the text that is being drawn...

Current patch:

image

v44r:
Ugh, maybe a bit ugly but it's the only solution I found. I find the baseline for ASCII text, find the baseline for the text that is going to be shown and shift it by the difference. Maybe there is a better way, but the documentation for pangocairo regarding vertical alignment is _a bit_ lacking.

image

image

At least the result seems visually correct.

Earnestly:
Just for reference since VTE was included, here is what xterm looks like:
xterm

v44r:
Doesn't seem to be a problem (full screen capture with 100 lines of text and -l100):

image

Cloudef:
>I find the baseline for ASCII text, find the baseline for the text that is going to be shown and shift it by the difference. Maybe there is a better way, but the documentation for pangocairo regarding vertical alignment is a bit lacking. That actually sounds like a really good solution to me. I also agree on the documentation. The screenshots also look great!

PR #46, opened 2019-03-24
Add options for renderers, make deps required

DanySpin97:


lib/renderers/CMakeLists.txt
```diff
@@ -1,8 +1,21 @@
...
+    SET(RENDERER_COUNT 1)
+    LIST(APPEND RENDERERS "x11")
+ENDIF()
+
+IF(BEMENU_WAYLAND_RENDERER)
+    SET(RENDERER_COUNT 1)
+    LIST(APPEND RENDERERS "wayland")
+ENDIF()
+
+IF(NOT RENDERER_COUNT)
```
Cloudef: cmake has no way to check if list is empty?

Cloudef:
The readme probably should be updated to reflect this change. I like this change otherwise.

CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ SET(BEMENU_DESCRIPTION "Dynamic menu library and client program inspired by dmen
 SET(BEMENU_VERSION "0.1.0")
 SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/CMake)
 
+OPTION(BEMENU_CURSES_RENDERER "Build curses backend" ON)
+OPTION(BEMENU_X11_RENDERER "Build X11 backend" ON)
+OPTION(BEMENU_WAYLAND_RENDERER "Build wayland renderer" OFF)
```
Cloudef: Instead of options, could also use single variable to customize the renderers, `BEMENU_RENDERERS` or something with default value `"curses;x11"` it could then be modified from command line with `-DBEMENU_RENDERERS="curses;x11;wayland"` for example.

DanySpin97: I don't like this because of how source-based distro packaging works; it is easier to assign a variable to an option/USE instead of using a list.

In exherbo packages format (called exheres) it would as simple as:

```bash
CMAKE_SRC_CONFIGURE_OPTIONS=(
    'ncurses BEMENU_CURSES_RENDERER'
    'X BEMENU_X11_RENDERER'
    'wayland BEMENU_WAYLAND_RENDERER'
)
```

While a `BEMENU_RENDERERS` would require creating a list in bash and appending the renderers.

lib/renderers/x11/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-FIND_PACKAGE(X11)
-FIND_PACKAGE(Cairo)
+FIND_PACKAGE(X11 REQUIRED)
+FIND_PACKAGE(Cairo REQUIRED)
 
 if (X11_FOUND AND X11_Xinerama_FOUND AND CAIRO_FOUND AND PANGO_FOUND)
```
Cloudef: this if should also be removed I guess

CMakeLists.txt
Cloudef: Fair enough.

Cloudef:
Ready to merge?

DanySpin97:
Fine by me.

PR #40, opened 2019-02-26
Fix blocking on Wayland

aidanharris:
Sometimes bemenu blocks in epoll_wait which prevents bemenu from showing
up until another key is pressed or the mouse is moved. This may be
working around a compositor specific bug in Sway/Wlroots.

Backtrace from gdb attached to a running bemenu process that failed to render because it was stuck in epoll_wait:

```c
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
0x00007f2f8a6156ab in epoll_wait (epfd=5, events=events@entry=0x7ffdf6422d20, maxevents=maxevents@entry=16, timeout=timeout@entry=-1) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
30	../sysdeps/unix/sysv/linux/epoll_wait.c: No such file or directory.

Thread 1 (Thread 0x7f2f8a505740 (LWP 13002)):
#0  0x00007f2f8a6156ab in epoll_wait (epfd=5, events=events@entry=0x7ffdf6422d20, maxevents=maxevents@entry=16, timeout=timeout@entry=-1) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
        resultvar = 18446744073709551612
        sc_ret = <optimized out>
#1  0x00007f2f8a73adc9 in render (menu=0x55f797907110) at /usr/src/debug/dev-libs/bemenu-9999/bemenu-9999/lib/renderers/wayland/wayland.c:29
        wayland = 0x55f7978ef9a0
[...]

PR #47, opened 2019-03-24
wayland: implement a proper repaint cycle

### LINKED ISSUE #42: Do not wait for initial wayland events when starting up
I have tried running bemenu on Wayfire (a wlroots based compositor which supports layer-shell). However the `bemenu-run` didn't do anything visible. After a bit of debugging, I think I found the reason: the wayland backend waits for a wayland event before drawing the first frame. In some compositors though no such event will happen before the surface is mapped. Thus I think something along the lines should be implemented:

```
diff --git a/lib/renderers/wayland/wayland.c b/lib/renderers/wayland/wayland.c
index dd97009..8c18b62 100644
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -13,6 +13,7 @@
 #include <sys/timerfd.h>
 
 static int efd;
+static int firstLoop = 1;
 
 static void
 render(const struct bm_menu *menu)
@@ -26,6 +27,16 @@ render(const struct bm_menu *menu)
     }
 
     struct epoll_event ep[16];
+    if (firstLoop)
+    {
+        struct window *window;
+        wl_list_for_each(window, &wayland->windows, link) {
+            bm_wl_window_render(window, wayland->display, menu);
+        }
+        firstLoop = 0;
+        return;
+    }
+
     uint32_t num = epoll_wait(efd, ep, 16, -1);
     for (uint32_t i = 0; i < num; ++i) {
         if (ep[i].data.ptr == &wayland->fds.display) {
```

This is of course just a hack, I am not familiar with the codebase enough to write a proper solution.

ammen99:
Fixes #42 
Supersedes #40 

With this change, bemenu now works correctly on Sway, Rootston and Wayfire, and key repeat isn't broken. 

How wayland rendering is done now: we have the `render_pending` flag to know when a window needs to be redrawn. Initially this is true, as we want to map the surface by attaching a buffer to it. When the user types or key repeat fires, we request a frame callback, set the `render_pending` flag when the callback is fired and repaint the surface(s) when control returns to the main loop.

### DIFF
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -25,6 +25,13 @@ render(const struct bm_menu *menu)
         return;
     }
 
+    struct window *window;
+    wl_list_for_each(window, &wayland->windows, link) {
+        if (window->render_pending)
+            bm_wl_window_render(window, wayland->display, menu);
+    }
+    wl_display_flush(wayland->display);
+
     struct epoll_event ep[16];
     uint32_t num = epoll_wait(efd, ep, 16, -1);
     for (uint32_t i = 0; i < num; ++i) {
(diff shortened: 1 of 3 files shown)


Cloudef:
Looks clean. Thanks.

Issue #45, opened 2019-03-19
bemenu is shown behind waybar


oblique:
bemenu is show behind waybar:
![screenshot_2019-03-19_21:21:59 697946539](https://user-images.githubusercontent.com/194040/54635635-50536e80-4a8d-11e9-8272-af2b6a904573.png)

I'm using the default configuration of [waybar](https://github.com/Alexays/Waybar).

Let me know if you need anything else.

Cloudef:
this is a compositor specific issue (or issue with the protocol waybar/bemenu uses), bemenu can't control where and how it appears under wayland, apart from what the protocol it uses allows.

brunelli:
Replacing `"layer": "top"` with `"layer": "bottom"` in waybar's config file solves this.

oblique:
I can confirm this. Thanks!

ammen99:
I think a proper solution would be changing how bemenu interacts with panel surfaces. We can tell the compositor to move the menu window just below the panel(s) by setting exclusive zone to 0.

 Cloudef would you merge such a PR, as this will slightly change the behavior of bemenu?

YorozuyaDev:
set layer to bot in waybar works but icon tray menu get overlapped by windows, is there a way to set bemenu overlap waybar?

oblique:
My current solution is `"layer": "top"` in waybar config and `-n` argument for bemenu.

YorozuyaDev:
> 
> 
> My current solution is `"layer": "top"` in waybar config and `-n` argument for bemenu.

but this only put bemenu over waybar

PR #48, opened 2019-03-25
Add option to respect panel position

ammen99:
Reading #45 I realized there is another way to solve the issue, and this time it is in bemenu. In addition, this is the behavior I actually want (perhaps others too). The command line option is `-n / --no-overlap`.

Screenshot of how bemenu behaves with `-n`:
screenshot-2019-03-25-21:56:37

Note that without this option, behavior is as it was before.


Cloudef:
squash the commits and it's good to go.

Cloudef:
realized there's squash and merge option

Issue #44, opened 2019-03-19
Allow specifying height


Mel34:
`--height <height>
           Sets the height of the bar. Default height (0) will match the font size.`

Reason for this would be matching the height of sway-bar, for us OCD people out there. :)

PR #57, opened 2019-05-31
bemenu: add --line-height / -H option

alyssais:
Text is displayed vertically centered in a line. If unspecified, or 0, the previous behaviour of making the height the size of the text, plus two pixels on either side, is used, so there will be no change in behaviour if this option is not used.

Haven’t tested in X because I don’t use it, but don’t see any reason it should be different to Wayland.

Fixes https://github.com/Cloudef/bemenu/issues/44.


alyssais:
The `-H` form doesn’t seem to work. Would appreciate help figuring out why.

alyssais:
Figured it out :)

Cloudef:
Thanks

PR #60, opened 2019-06-05
Prevent a deadloop in PATH handling

### LINKED ISSUE #59: Empty path segments causing bemenu-run to go in a deadloop
An empty PATH segment ("::") is causing bemenu-run to go into a deadloop.

**To reproduce**
```sh
PATH="$PATH::" ./client/bemenu-run
```

**Expected Behaviour**

bemenu-run should ignore empty path segments

toke:
Empty PATH segments should no longer cause
a deadloop.

Fixes #59


Cloudef:
Thanks

PR #61, opened 2019-07-12
Add C-g shortcut to x11 and wayland

sauyon:
Very simple case to add the C-g shortcut I normally use to kill dmenu.


Cloudef:
Thank you

PR #64, opened 2019-08-07
Add man pages for bemenu binaries

khardix:
The man page(s) contents are mostly just more verbosely re-written help messages. Please check for any factual mistakes I might have made.

---

Should close Cloudef#27 if/when accepted.

### DIFF
--- a/man/bemenu.1
+++ b/man/bemenu.1
@@ -39,7 +39,7 @@ Entering text will narrow the items to those matching the tokens in the input.
 is a special-case invocation of
 .BR bemenu ,
 where the input is a list of executables in the $PATH directories,
-and the selection is executed by the $SHELL.
+and the selection gets executed.
 
 .SH OPTIONS
 .TP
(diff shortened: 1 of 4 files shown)


man/bemenu.1
```diff
@@ -0,0 +1,194 @@
...
+It reads a list of newline-separated items from standard input.
+When the user selects an item and presses Return,
+their choice is printed to standard output and bemenu terminates.
+Entering text will narrow the items to those matching the tokens in the input.
+
+.B bemenu-run
+is a special-case invocation of
+.BR bemenu ,
+where the input is a list of executables in the $PATH directories,
+and the selection is executed by the $SHELL.
```
Cloudef: the bemenu-run doesn't actually use $SHELL, instead it forks and execs.

khardix: Corrected. I took too much inspiration from the `dmenu` man page without fact checking it :sweat_smile:

Cloudef:
Thanks! Merged

PR #70, opened 2019-11-05
Handle HiDPI scaling on Wayland

### LINKED ISSUE #38: Blurry font rendering while using HiDPI scaling on wayland
I have issues with the bemenu font rendering when HiDPI scaling (factor 2) is used on wayland. I'm using sway (version 1.0-beta.2-270-g1a1133dc) and bemenu (r262.799d8b9-1 according to the package version I installed from git) on Archlinux.
I also tried to set the backend specifically to wayland using the `BEMENU_BACKEND` env variable and using different ttf fonts and sizes. But with no success. The font appears very blurry in comparison to other applications.

lheckemann:
A PR with @hexd0t's hidpi implementation. Is X support a requirement for this to be merged?

Fixes #38.

### DIFF
--- a/lib/renderers/cairo.h
+++ b/lib/renderers/cairo.h
@@ -12,6 +12,7 @@ struct cairo {
     cairo_t *cr;
     cairo_surface_t *surface;
     PangoContext *pango;
+    int scale;
 };
 
 struct cairo_color {
@@ -129,6 +130,9 @@ bm_cairo_draw_line(struct cairo *cairo, struct cairo_paint *paint, struct cairo_
     if (!ret)
         return false;
 
+
+    cairo_scale(cairo->cr, cairo->scale, cairo->scale);
+
     PangoLayout *layout = bm_pango_get_layout(cairo, paint, buffer);
     pango_cairo_update_layout(cairo->cr, layout);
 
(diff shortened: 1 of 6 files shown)


Cloudef:
>Is X support a requirement for this to be merged?

No, but would be nice. Somebody else can do it if interested.

lib/renderers/wayland/registry.c
```diff
@@ -243,7 +243,10 @@ display_handle_done(void *data, struct wl_output *wl_output)
 static void
 display_handle_scale(void *data, struct wl_output *wl_output, int32_t scale)
 {
-    (void)data, (void)wl_output, (void)scale;
+    (void)wl_output;
+    struct output *output = data;
+
+    output->scale = scale;
```
Cloudef: Can the scale here be 0?

lib/renderers/cairo.h
```diff
@@ -129,6 +130,9 @@ bm_cairo_draw_line(struct cairo *cairo, struct cairo_paint *paint, struct cairo_
     if (!ret)
         return false;
 
+
```
Cloudef: `assert(cairo->scale > 0);` here would be nice addition

Cloudef:
Thank you

PR #78, opened 2020-01-08
Add Ctrl-[ shortcut to x11 and wayland

sleich:
Ctrl-[ shortcut is used to quit bemenu. It already works with ncurses backend


Cloudef:
Thanks

Issue #79, opened 2020-01-09
bemenu suddenly stopped working


AdrienLemaire:
bemenu worked fine until yesterday, but suddenly doesn't do anything anymore.
```console
$ ls | bemenu

$
```

I cannot find any debug option in the bemenu manpages, so I have no idea what's happening and how to fix the issue.

```console
$ bemenu --version                                           │
bemenu v0.3.0 
$ pacman -Qs bemenu
local/bemenu 0.3.0-1
    Dynamic menu library and client program inspired by dmenu
    with support for wayland compositors
```
Running arch linux and sway 1:1.2-5 (wayland).

Any idea why this is happening?

Cloudef:
Maybe #76 is related. Bisecting bemenu or sway can help here.
https://git-scm.com/docs/git-bisect

zimbipw:
Having the same issue here. Downgrading to bemenu 0.1. brings back the menu.
pacman -Ss bemenu
community/bemenu 0.3.0-1 [Installiert: 0.1.0-1]

The temporary fix from https://github.com/Cloudef/bemenu/issues/76 using swaymsg reload dosn't work for me though.

depate:
This is related to this [bug](https://bugs.archlinux.org/index.php?do=details&task_id=65087).

Seems to be that the wayland renderer was not compiled into the current package. 

#76 is most likely no related.

escondida:
Actually, it's not just that -DCMAKE_OPTION, apparently. When building in a clean chroot (with wayland installed, naturally), even adding -DBEMENU_WAYLAND_RENDERER=ON to cmake's options doesn't help: it builds and installs the necessary .so, but when testing on weston it silently fails.

It's not a silent dependency on egl-wayland (though cmake does check for its presence). but since depate got it working when building (I assume) outside of a clean chroot, there's likely some kind of undocumented dependency thing going on here.

I guess I'll investigate this more tomorrow.

escondida:
P.S.: Again, I'm testing in weston, since it's the reference implementation of a wayland compositor. If you want to build in a clean chroot on Arch when testing changes, use extra-x86_64-build from devtools.

depate:
@escondida 
I'm no newbie to ArchLinux, but never had to deal with proper packaging or even package building. You're right: I did not use chroot. I have to read the [guides](https://wiki.archlinux.org/index.php/DeveloperWiki:Building_in_a_clean_chroot) first. But I'm already working on it.

[...]

PR #84, opened 2020-01-29
Document that bemenu is works only with wlroots-based compositors

escondida:
See https://github.com/Cloudef/bemenu/issues/79#issuecomment-572867783


Cloudef:
Merged, thanks

PR #86, opened 2020-02-06
Mark global wayland constant extern

khardix:
Without `extern`, the changed lines are not declarations, but "tentative definitions" (according to GCC man page, option `-fcommon`). When specified in a header file that is included in more than one `.c` file, these result in linking failure unless `-fcommon` is specified.

GCC 10 changed the default from `-fcommon` to `-fno-common`, and as such the previous code no longer links properly.

With `extern`, these lines are considered declarations, and the linking proceeds successfully.

### DIFF
--- a/lib/renderers/wayland/wayland.h
+++ b/lib/renderers/wayland/wayland.h
@@ -33,8 +33,8 @@ enum mask {
     MASK_LAST
 };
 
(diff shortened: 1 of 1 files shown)


Cloudef:
Thanks, merged

PR #91, opened 2020-02-09
Don't attempt to chmod everything under PREFIX

### LINKED ISSUE #90: make install PREFIX=/usr screws sudo permissions
This line: https://github.com/Cloudef/bemenu/blob/master/GNUmakefile#L101
sets `sudo` permissions to `0775` which renders it useless because `sudo` permissions should be `4775`

Earnestly:
Fixes #90

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -98,7 +98,8 @@ install-renderers:
 install-bins:
 	mkdir -p "$(DESTDIR)$(PREFIX)$(bindir)"
 	-cp $(bins) "$(DESTDIR)$(PREFIX)$(bindir)"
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -98,7 +98,8 @@ install-renderers:
 install-bins:
 	mkdir -p "$(DESTDIR)$(PREFIX)$(bindir)"
 	-cp $(bins) "$(DESTDIR)$(PREFIX)$(bindir)"
-	-chmod 0755 "$(DESTDIR)$(PREFIX)$(bindir)"/*
+	-chmod 0755 $(addprefix "$(DESTDIR)$(PREFIX)$(bindir)"/,$(bins))
+
```
Earnestly: Sigh the github editor is terrible.

PR #94, opened 2020-02-22
Add bemenu_path script

electrickite:
dmenu is generally packaged with a simple shell script, `dmenu_path` that generates a list of all files within a user's `$PATH`. This PR adds an equivalent script, `bemenu_path`, that does the same. This allows for use cases that use bemenu to select a program from the path, but not execute it.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -19,6 +19,7 @@ override CPPFLAGS += -D_DEFAULT_SOURCE -Ilib
 libs = libbemenu.so
 pkgconfigs = bemenu.pc
 bins = bemenu bemenu-run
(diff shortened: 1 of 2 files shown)


Cloudef:
stest seems to be part of dmenu, alternative needs to be either provided, or you might as well use dmenu_path. Note there are also alternatives to stest, even shell builtins if you use non posix shell. These shells cache the commands already internally, so you don't need a separate cache file.

electrickite:
Hmm, good point. Relying on a dmenu provided utility defeats the purpose, doesn’t it :)

Maybe another approach would be better. I’ll close this one.

PR #95, opened 2020-02-22
Add --no-exec option to bemenu-run

electrickite:
As an alternative to the `dmenu_path` script, this PR adds a `--no-exec` option to `bemenu-run`. Setting this option will cause `bemenu-run` to print the selected command name to stdout rather than execute it.


Cloudef:
Thanks, merged

Issue #71, opened 2019-11-08
Transparency


mb720:
Hi and thanks a lot for bemenu!

I was wondering if there's a way to specify an alpha value in bemenu's colors.

For example: `bemenu --nb '#09abac5a'` where the last two characters would make the background transparent.

Cloudef:
I think this could be possible yeah.

Cloudef:
This is now implemented.

PR #96, opened 2020-02-27
Add transparency support

vilhalmer:
This was a much smaller diff than anticipated! Colors are now accepted with or without a fourth pair of hex values for alpha. Without alpha specified, it defaults to FF as you would expect.

### DIFF
--- a/lib/menu.c
+++ b/lib/menu.c
@@ -15,18 +15,18 @@ static const char *default_font = "monospace 10";
  * Default hexadecimal colors.
  */
 static const char *default_colors[BM_COLOR_LAST] = {
-    "#121212", // BM_COLOR_TITLE_BG
-    "#D81860", // BM_COLOR_TITLE_FG
-    "#121212", // BM_COLOR_FILTER_BG
-    "#CACACA", // BM_COLOR_FILTER_FG
-    "#121212", // BM_COLOR_ITEM_BG
-    "#CACACA", // BM_COLOR_ITEM_FG
-    "#121212", // BM_COLOR_HIGHLIGHTED_BG
-    "#D81860", // BM_COLOR_HIGHLIGHTED_FG
-    "#121212", // BM_COLOR_SELECTED_BG
-    "#D81860", // BM_COLOR_SELECTED_FG
-    "#121212", // BM_COLOR_SCROLLBAR_BG
-    "#D81860", // BM_COLOR_SCROLLBAR_FG
+    "#121212FF", // BM_COLOR_TITLE_BG
+    "#D81860FF", // BM_COLOR_TITLE_FG
+    "#121212FF", // BM_COLOR_FILTER_BG
+    "#CACACAFF", // BM_COLOR_FILTER_FG
+    "#121212FF", // BM_COLOR_ITEM_BG
+    "#CACACAFF", // BM_COLOR_ITEM_FG
+    "#121212FF", // BM_COLOR_HIGHLIGHTED_BG
+    "#D81860FF", // BM_COLOR_HIGHLIGHTED_FG
+    "#121212FF", // BM_COLOR_SELECTED_BG
+    "#D81860FF", // BM_COLOR_SELECTED_FG
+    "#121212FF", // BM_COLOR_SCROLLBAR_BG
+    "#D81860FF", // BM_COLOR_SCROLLBAR_FG
 };
 
 /**
@@ -301,8 +301,9 @@ bm_menu_set_color(struct bm_menu *menu, enum bm_color color, const char *hex)
 
     const char *nhex = (hex ? hex : default_colors[color]);
 
-    unsigned int r, g, b;
-    if (sscanf(nhex,"#%2x%2x%2x", &r, &b, &g) != 3)
+    unsigned int r, g, b, a = 255;
+    int matched = sscanf(nhex, "#%2x%2x%2x%2x", &r, &b, &g, &a);
+    if (matched != 3 && matched != 4)
         return false;
 
     char *copy = NULL;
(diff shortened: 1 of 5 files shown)


vilhalmer:
This fixes #71.

Cloudef:
Very nice! Did you test this on which platform? X11 might need RGBA bits set on window creation. Wayland i think would work without any changes.

vilhalmer:
Just fixed up clearing the cairo surface, missed pushing that up. I've only tested on wayland, I can take a look at X tomorrow morning.

Cloudef:
If you are willing to do X support as well. I'll wait until that before merging this. Looks good.

Cloudef:
Also note that X11 most likely won't work without a compositor. Fake transparency could be done for non compositor support. But that could as very well be another feature, another pr.

lib/menu.c
```diff
@@ -301,8 +301,9 @@ bm_menu_set_color(struct bm_menu *menu, enum bm_color color, const char *hex)
 
     const char *nhex = (hex ? hex : default_colors[color]);
 
-    unsigned int r, g, b;
-    if (sscanf(nhex,"#%2x%2x%2x", &r, &b, &g) != 3)
+    unsigned int r, g, b, a = 255;
+    int matched = sscanf(nhex, "#%2x%2x%2x%2x", &r, &b, &g, &a);
+    if (matched != 3 && matched != 4)
```
panaman67:
```suggestion
    if (matched > 2)
```

Cloudef: That would make it accept color codes with more than 3 or 4 components, so not sure if that's good idea 🤔

vilhalmer: I'm fine with either way, but only checking < 3 will allow extra characters to slip by.

vilhalmer:
>If you are willing to do X support as well

I have no experience with Xlib, so if you'd rather handle that I will defer to you. But I'm also willing to learn a bit. :)

Cloudef:
https://gist.github.com/je-so/903479/834dfd78705b16ec5f7bbd10925980ace4049e17#file-testprogram-c-L53-L55
Basically these are the only bits you should need in bm_x11_window_create. Fallback to defaultvisual and defaultdepth if creating window with 32 bit visual and depth fails.

Cloudef:
I think if you are testing in XWayland, that should already be composed (by the wayland composer), so RGBA should work fine there.

vilhalmer:
I've pushed what I believe to be the correct solution. However, I'm unable to test properly because even current master refuses to start for me in Xwayland with `BEMENU_BACKEND=x11`. :(

I was able to test that it still runs correctly under native X, but I don't have an X compositor installed to see if it really becomes transparent.

Cloudef:
Compton is something you could use to test the X11 side quickly. https://github.com/chjj/compton

Earnestly:
Now developed in this fork: https://github.com/yshui/picom

Cloudef:
Tested it! Seems to work.

Cloudef:
if I do `bemenu-run -l 15 -nb #00000000` and go down the list, it acts bit strangely. I guess there are no clears before redrawing?

Cloudef:
Also, the transparency seems to work even without a compositor. I guess either cairo or x11 does the fake transparency for you.

vilhalmer:
>I guess there are no clears before redrawing?

Hmm, I thought this was fixed in ffd773f. Maybe cairo behaves differently when using the X backend?

vilhalmer:
I'm not sure that dc33df1 fixes the redrawing issue on X, but it makes more sense to use CAIRO_OPERATOR_CLEAR to clear either way. :)

vilhalmer:
The good news is I got transparency to work in X, not sure what I was doing wrong before. The bad news is that I've been bashing my head against `bm_x11_window_render` for a while and I have no idea what is going wrong. It seems like the cairo group is being used instead of a real double buffer, which I'm not sure works with transparency? Any change I make to the function either does nothing or eliminates the transparency.

Cloudef:
Why do you need to edit `bm_x11_window_render`?

Cloudef:
Ah, I just tested your new code. I see the issue now. Lemme send you a patch that should fix it in a moment.

Cloudef:
```
diff --git a/lib/renderers/x11/window.c b/lib/renderers/x11/window.c
index bca1003..daf129a 100644
--- a/lib/renderers/x11/window.c
+++ b/lib/renderers/x11/window.c
@@ -94,8 +94,11 @@ bm_x11_window_render(struct window *window, const struct bm_menu *menu)
     }
 
     if (buffer->created) {
+        cairo_save(buffer->cairo.cr);
+        cairo_set_operator(buffer->cairo.cr, CAIRO_OPERATOR_SOURCE);
         cairo_paint(buffer->cairo.cr);
         cairo_surface_flush(buffer->cairo.surface);
+        cairo_restore(buffer->cairo.cr);
     }
 }
```

This should do it. It will only work with compositor. Doing pseudo-transparency would involve using XCopyArea or other tricks. But that would be separate thing.

Cloudef:
Note with latest diff. Not sure if the clearing and such is needed in cairo render, but maybe they are still for wayland?

vilhalmer:
(Sorry, have been busy with other things. Will get back to this soon!)

vilhalmer:
Alright! With your patch and picom running, looks good on X. I think this is ready to go. Do you want me to squash things into a nicer set of commits, or will you handle that when merging?

Cloudef:
Lets see how github's "squash and merge" works

Cloudef:
Merged, thank you!

vilhalmer:
Thanks for your help!

PR #99, opened 2020-03-15
add soname to shared libraries

prometheanfire:
Found in Gentoo QA

* QA Notice: The following shared libraries lack a SONAME
* /usr/lib/libbemenu.so.0.4.0

Signed-off-by: Matthew Thode <mthode@mthode.org>


Cloudef:
Thank you!

PR #101, opened 2020-03-19
x11,wayland: treat keypad enter as normal enter

cole-h:
Sometimes I'm so lazy that I don't want to reach for the actual enter key on my keyboard, and hit the enter key on my keypad, but that doesn't work. This adds the keypad return key and sets its functionality to be that of a normal enter key.

These changes were really just copy-paste (and some keysym lookups). Let me know if there's something I missed, or something needs to be changed. I can't test this on x11, so I'd appreciate any help in that department.

### DIFF
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -118,6 +118,7 @@ poll_key(const struct bm_menu *menu, unsigned int *unicode)
         case XKB_KEY_Insert:
             return BM_KEY_SHIFT_RETURN;
 
+        case XKB_KEY_KP_Enter:
         case XKB_KEY_Return:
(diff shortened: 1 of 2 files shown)


lib/bemenu.h
```diff
@@ -190,6 +190,7 @@ enum bm_key {
     BM_KEY_SHIFT_TAB,
     BM_KEY_ESCAPE,
     BM_KEY_RETURN,
+    BM_KEY_KP_RETURN,
```
Cloudef: No need for this. Map the numpad return to BM_KEY_RETURN in wayland and x11 renderer.

Cloudef: The keys here are only for unique actions. Since _RETURN and _KP_RETURN do the same action, we don't need separate enum for it.

cole-h: Thanks!

Cloudef:
Thanks, merging!

PR #112, opened 2020-04-18
unicode fixes

siiptuo:


Cloudef:
Can you describe what the problem here is?

lib/renderers/cairo.h
```diff
@@ -155,11 +155,8 @@ bm_cairo_draw_line(struct cairo *cairo, struct cairo_paint *paint, struct cairo_
     pango_cairo_show_layout(cairo->cr, layout);
 
     if (paint->draw_cursor) {
-        size_t chr = 0;
-        for (size_t c = 0; c < paint->cursor; ++chr, c += bm_utf8_rune_next(buffer, c));
-
         PangoRectangle rect;
-        pango_layout_index_to_pos(layout, chr, &rect);
+        pango_layout_index_to_pos(layout, paint->cursor, &rect);
```
Cloudef, line 160: This most likely will break multibyte text. (CJK)

Cloudef: Note that chr is byte index.

lib/menu.c
```diff
@@ -710,8 +710,11 @@ bm_menu_run_with_key(struct bm_menu *menu, enum bm_key key, uint32_t unicode)
             break;
 
         case BM_KEY_DELETE:
-            if (menu->filter)
-                bm_utf8_rune_remove(menu->filter, menu->cursor + 1, NULL);
+            if (menu->filter) {
+                size_t width = bm_utf8_rune_next(menu->filter, menu->cursor);
+                if (width)
+                    bm_utf8_rune_remove(menu->filter, menu->cursor + width, NULL);
+            }
```
Cloudef, line 717: This looks correct :+1:

lib/renderers/cairo.h
Cloudef: Or isn't, I can't read. Weird, are you sure this takes byte index? The documentation seems to talk about "grapheme".
>Converts from an index within a PangoLayout to the onscreen position corresponding to the grapheme at that index

Cloudef: >index_ | byte index within layout

Seems so.. I remember this not working with simply `paint->cursor` though :S
Can you test with some CJK text?

siiptuo:
> Can you describe what the problem here is?

Trying to fix issues I noticed after typing something like "äää" and moving the cursor around.

lib/renderers/cairo.h
siiptuo: CJK text seems to work. I don't have a good way of typing these characters so I tested with `bm_menu_set_filter` and `xdotool type`.

Cloudef:
Okay, merging this. The changes seem to make sense.

Cloudef:
Thanks!

PR #116, opened 2020-05-07
GNUmakefile: add individual install rules for each renderer

escondida:
This eases split packaging as suggested in
https://github.com/Cloudef/bemenu/issues/79#issuecomment-572867783

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -100,10 +100,6 @@ install-libs: $(libs)
 install-lib-symlinks: $(libs) | install-libs
 	cp -RP $^ $(addsuffix .$(firstword $(subst ., ,$(VERSION))), $^) "$(DESTDIR)$(PREFIX)$(libdir)"
 
-install-renderers:
-	mkdir -p "$(DESTDIR)$(PREFIX)$(libdir)/bemenu"
-	-cp $(renderers) "$(DESTDIR)$(PREFIX)$(libdir)/bemenu"
-
 install-bins:
 	mkdir -p "$(DESTDIR)$(PREFIX)$(bindir)"
 	-cp $(bins) "$(DESTDIR)$(PREFIX)$(bindir)"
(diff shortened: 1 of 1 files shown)


escondida:
By the way, great idea switching to plain Make!

Not directly related to this commit, but something I noticed on the way: is there a reason why libbemenu.so and its links aren't also installed to /usr/lib/bemenu?

Cloudef:
>Not directly related to this commit, but something I noticed on the way: is there a reason why libbemenu.so and its links aren't also installed to /usr/lib/bemenu?

This is because `libbemenu` is the library the clients use. Normal libraries are usually directly in `/usr/lib/`

Cloudef:
Btw, you can already do the same, by doing:
```
make x11
make install
```
This would only install x11 component.

escondida:
re: lib location, fair enough!

As to the main patch, my use case is packaging for Arch, in which the source is built once, and then (in the case of a split package, such as you suggested bemenu be), an install function is run for each subpackage using the built source. I'd imagine it works pretty similarly in other packaging situations, too (-:

escondida:
You can see what I mean by looking at the current [PKGBUILD](https://git.archlinux.org/svntogit/community.git/plain/trunk/PKGBUILD?h=packages/bemenu), in which I have split the package, but have to manually remove unwanted components in each install phase.

Cloudef:
Fair enough. Merging, thanks!

GNUmakefile
```diff
@@ -113,7 +109,23 @@ install-man: man/bemenu.1 man/bemenu-run.1
 	mkdir -p "$(DESTDIR)$(PREFIX)$(mandir)"
 	cp $^ "$(DESTDIR)$(PREFIX)$(mandir)"
 
-install: install-pkgconfig install-include install-lib-symlinks install-renderers install-bins install-man
+install-renderers: install-curses install-wayland install-x11
+
+install-curses: curses
+	install -d "$(DESTDIR)$(PREFIX)$(libdir)/bemenu"
```
Earnestly, line 115: These should probably be `mkdir -p` and `cp` as is the case elsewhere.  However, while `install` is not POSIX, it appears that `install -d`  and `install` are both "de facto portable" across Linux, BSD and MacOS.

Cloudef: I changed these already in separate commits after noticing :)

Issue #126, opened 2020-07-11
Custom actions using keybinding


SevereOverfl0w:
This is a feature idea from Rofi, which I'm trying to replicate under Wayland with Bemenu.

Rofi will vary the exit code based on whether the user used on of the `kb-custom` bindings or not.  These are each bound to Alt+[1-9] and some additional ones past that (e.g. kb-custom-12 is Alt-@).

The use-case is to enable things like open/remove on a list of files, or edit/do/remove on a todo list.  When pressing Alt-1 on an item, the process exists as if hitting enter, but the exit code deviates based on that.

PR #127, opened 2020-08-03
Add custom key support

SevereOverfl0w:
Implementation of #126.  Currently in draft & only implemented for
Wayland (what I primarily use).

I wanted to get some early feedback on the implementation.  I haven't
done C for many years, and I want to make sure my approach is in keeping
with how you'd like to see this done before continuing.


SevereOverfl0w:
ncurses and x11 implementations added.  The ncurses one is a little bit of a hack, but supports the <esc> prefix version mentioned here: https://stackoverflow.com/a/10216459.  The 8-bit encoding currently sends other characters, I can't find information on whether this can be detected or not.  I can extend support to detect those cases, but if users expect their terminal to send certain characters on alt-X then it will not work for those cases.  Most of the extended ascii keys aren't ones you would press normally anyway though.

Where would you like this functionality to be documented?

SevereOverfl0w:
I've extended support to include high bit by default, branch can easily be deleted if it's not suited, etc.

Cloudef:
Looks good. If the ncurses thing works in xterm, it's good baseline to know whether it would work on other terminal usually.

SevereOverfl0w:
xterm only works with the high bit code (by default, this can be configured in xterm).

Kitty, the other terminal I tried, sent <esc> followed by 1.

Cloudef:
If the terminal support seems spotty. I would be fine having this feature only for the GUIs.

SevereOverfl0w:
I think these two methods capture the two methods that exist, or at least the two major ones.

As far as I can tell, neovim works by checking these two methods (based on docs and bug reports where it was extended/fixed).

SevereOverfl0w:
Just to clarify, are you waiting on me to remove the terminal code?

Cloudef:
If you think it works with majority of terminal emulators. I'm fine with having the code in. I'll test and merge this later.

SevereOverfl0w:
Great, thanks! I'm looking forward to using this on all my machines.

Cloudef:
Sorry for the late response. I tested this with iTerm on OSX since I don't have other machine available right now. It doesn't seem to work there, but it doesn't seem to break functionality either so it's fine. I assume it will work with the more popular terminal emulators though (especially under linux) so I'm fine with this.

SevereOverfl0w:
Apparently iterm2 doesn't do this by default https://iterm2.com/faq.html but I'm not sure what it does send by default.

Libtickit is probably the best way to get full coverage for this stuff.

SevereOverfl0w:
Actually, it goes a step further and implements a libtickit specific solution (which neovim uses under the hood)

PR #135, opened 2020-10-09
Fix #98

### LINKED ISSUE #98: keypresses lost/eaten occasionally
If I open bemenu-run and type stuff, I lose two letters. I don't know the exact interval, and it's not always the same letters. I tried to type "the quick brown fox jumps over the lazy dog" two times. The first time I lost the "wn" in "brown". The second time I think I lost part of the word "jumps" so it said "jus". It's the sorta thing that could drive a man crazy.
I'm on version 0.3.0 from the NixOS repos, so apologies if this bug has already been fixed in a later version. I searched the issues for "keypress" and saw nothing.
I believe this happens with bemenu in general, by the way. I also have a clipboard script that uses it and seems to lose keypresses, so it's not specific to the run command.
I'm using Sway (wayland), if that matters.

fancycade:
Add key_pending field to sync wayland keyboard event loops

This fixes #98 

I have tested this out on a wayland environment and can confirm it works well. The basic idea is that the state of input.sym changes with key release only when the poll_key method has already used the value.

### DIFF
--- a/lib/renderers/wayland/registry.c
+++ b/lib/renderers/wayland/registry.c
@@ -107,8 +107,15 @@ keyboard_handle_leave(void *data, struct wl_keyboard *keyboard, uint32_t serial,
 static void
 press(struct input *input, xkb_keysym_t sym, uint32_t key, enum wl_keyboard_key_state state)
 {
-    input->sym = (state == WL_KEYBOARD_KEY_STATE_PRESSED ? sym : XKB_KEY_NoSymbol);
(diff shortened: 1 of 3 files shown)


Cloudef:
Makes sense to me. Does key repeat work as well still?

fancycade:
Yes repeating keys still works. Repeat still works exactly like key press.

Cloudef:
Thanks, merging this.

Issue #102, opened 2020-03-24
bemenu is displayed only in one monitor


zwma:
Building from https://github.com/Cloudef/bemenu/commit/cd53b7bb555cf1c5afaae3779a88e126571faf8c, Wayland backend.

Now bemenu only appears in display marked as 0 by default. (In my case it's an external monitor). I can change it to 1 using the option `-m`, but there is no way (that I know) to have it show up in both displays. It used to be like that in 0.3.0

Cloudef:
https://github.com/Cloudef/bemenu/pull/103 Can you try if this PR works for you. You need to pass `-1` to `-m` option. This works only for wayland.

dwma:
Just tried it and it didn't work, adding `-m -1` behaves like not having `-m` set at all.

`0` and `1` works as before although this time is reversed (don't know if the numbers are assigned randomly or the change is related to this PR)

FWIW, the compiler complained about comparing `uint` and `int`.

Cloudef:
I think I messed up the if condition. I don't really have proper wayland setup here to test myself. Anyhow updated the PR, you can try again. As extra `all` option for `-m` should be equivalent to `-1` now too.

dwma:
No complaints from the compiler this time but the end result is exactly the same :( I tried both `-1` and `all`.

Cloudef:
Edited the pr again :)

dwma:
No luck, same as before.

Cloudef:
Dunno, the latest changes should work by only reading it. But since I don't have wayland setup I can't really test either. The PR could be worked on by someone else in the meantime.

Cloudef:
Ah, I did notice one error still at my edits. Updated again.

dwma:
Yep, worked this time!

Cloudef:
Did I break anything else?

dwma:
Tested the `-m` option with `0`, `1`, `-1` and `all`  and it's working fine.

However, now you can't keep pressed a key (backspace to delete, for example). This worked in https://github.com/Cloudef/bemenu/commit/cd53b7bb555cf1c5afaae3779a88e126571faf8c

Cloudef:
Do you mean key repeat?

dwma:
Yes, exactly.

Cloudef:
For key repeat, I'm not aware of any recent commits that would've changed the behavior for wayland. If you can you should do a git bisect to find the commit that broke it.

dwma:
[...]

PR #136, opened 2020-10-14
Update wlr-layer-shell-unstable-v1.xml

fancycade:
I was hoping this could help out with #102 . I simply copied the current version of this from [mako](https://github.com/emersion/mako).


Cloudef:
I think the proper fix is to do https://github.com/Cloudef/bemenu/issues/102#issuecomment-650778958 instead to fix #102.
By the way the mako one is also quite old, https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-layer-shell-unstable-v1.xml this is the place for most recent protocols.

fancycade:
Okay I'll circle back to this.

PR #141, opened 2020-11-14
Add --filter option.

### LINKED ISSUE #140: Add parameter for pre-selection of entries.
Hi,

first off: thank you so much for bemenu, I really enjoy using it. I was wondering if it is possible to pre-fill the input prompt to provide a default subset of entries (I didn't find a solution in the docs), or if you would consider (taking patches for) such a feature?

Rationale: I use bemenu in combination with the [pass](https://www.passwordstore.org/) password manager.
I would like to offer all entries that match a certain string as a pre-selection.
E.g., the following command:
```
bemenu --pre-selection "reddit.com"
"reddit.com/foobar"
"reddit.com/bar"
"gmail.com/foobar"
```
would pre-fill the input prompt with `reddit.com` and thus only show `reddit.com/foobar` and `reddit.com/bar`.
But the user could just delete everything and still choose from the whole set of all available entries.

jmehne:
Trying to close #140 - either I'm missing something, or this is actually this straight forward.

```
bemenu --filter foobar
foo
foobar
foobarbaz
^D
```

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -168,6 +168,7 @@ usage(FILE *out, const char *name)
           " -h, --help            display this help and exit.\n"
           " -v, --version         display version.\n"
           " -i, --ignorecase      match items case insensitively.\n"
+          " -F, --filter          filter entries for a given string.\n"
           " -w, --wrap            wraps cursor selection.\n"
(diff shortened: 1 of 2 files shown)


Cloudef:
Looks good, thanks.

PR #144, opened 2020-12-12
Declare filter_dmenu_fun as static

petercolberg:
This is an internal function that should not be exported from the library.


Cloudef:
Thank you

PR #145, opened 2020-12-13
Support compilation with -fvisibility=hidden

petercolberg:
When using a compiler with support for GNU C extensions (GCC, Clang),
explicitly mark functions shared across library boundaries as visible.

This is the default visibility and has no effect when compiling with the
default CFLAGS. When compiling with -fvisibility=hidden, however, this
exports only functions marked BM_PUBLIC and hides all others.

https://gcc.gnu.org/wiki/Visibility

This facilitates packaging for distributions that track shared library
symbols, e.g., Debian, which uses a symbol file to provide the minimal
version associated to each symbol exported by a library.

https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#the-symbols-system
https://manpages.debian.org/unstable/dpkg-dev/dpkg-gensymbols.1.en.html

### DIFF
--- a/lib/bemenu.h
+++ b/lib/bemenu.h
@@ -1,6 +1,22 @@
 #ifndef _BEMENU_H_
 #define _BEMENU_H_
 
+/*
+ * When using a compiler with support for GNU C extensions (GCC, Clang),
+ * explicitly mark functions shared across library boundaries as visible.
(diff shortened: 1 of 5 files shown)


Cloudef:
Thank you

PR #150, opened 2021-01-23
support hiding filter input

robertgzr:
"password mode"

Signed-off-by: Robert Günzler <r@gnzler.io>

### DIFF
--- a/lib/renderers/cairo.h
+++ b/lib/renderers/cairo.h
@@ -247,7 +247,15 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
     paint.cursor = menu->cursor;
     paint.pos = (struct pos){ (menu->title ? 2 : 0) + result.x_advance, vpadding };
     paint.box = (struct box){ (menu->title ? 2 : 4), 0, vpadding, vpadding, width - paint.pos.x, ascii_height };
-    bm_cairo_draw_line(cairo, &paint, &result, "%s", (menu->filter ? menu->filter : ""));
+
+    char *filter_text = (menu->filter ? menu->filter : "");
+    size_t filter_len = strlen(filter_text);
+    if (menu->password) {
+        memset(filter_text, '*', filter_len);
+        filter_text[filter_len] = '\0';
+    }
+
+    bm_cairo_draw_line(cairo, &paint, &result, "%s", filter_text);
     paint.draw_cursor = false;
     const uint32_t titleh = result.height;
     out_result->height = titleh;
--- a/lib/renderers/cairo.h
+++ b/lib/renderers/cairo.h
@@ -251,11 +251,11 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
     char *filter_text = (menu->filter ? menu->filter : "");
     size_t filter_len = strlen(filter_text);
     if (menu->password) {
-        memset(filter_text, '*', filter_len);
-        filter_text[filter_len] = '\0';
+        bm_cairo_draw_line(cairo, &paint, &result, "");
+    } else {
+        bm_cairo_draw_line(cairo, &paint, &result, "%s", filter_text);
     }
 
-    bm_cairo_draw_line(cairo, &paint, &result, "%s", filter_text);
     paint.draw_cursor = false;
     const uint32_t titleh = result.height;
     out_result->height = titleh;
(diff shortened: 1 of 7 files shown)


lib/renderers/cairo.h
```diff
@@ -247,7 +247,15 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
...
-    bm_cairo_draw_line(cairo, &paint, &result, "%s", (menu->filter ? menu->filter : ""));
+
+    char *filter_text = (menu->filter ? menu->filter : "");
+    size_t filter_len = strlen(filter_text);
+    if (menu->password) {
+        memset(filter_text, '*', filter_len);
+        filter_text[filter_len] = '\0';
+    }
+
+    bm_cairo_draw_line(cairo, &paint, &result, "%s", filter_text);
```
Cloudef, line 258: Instead if memsetting the filter_text, you could instead just draw empty string here if password is true.

lib/renderers/curses/curses.c
```diff
@@ -221,7 +221,14 @@ render(const struct bm_menu *menu)
...
 
-    draw_line(0, 0, "%*s%s", title_len, "", (menu->filter ? menu->filter + doffset : ""));
+    char *filter_text = (menu->filter ? menu->filter : "");
+    size_t filter_len = strlen(filter_text);
+    if (menu->password) {
+        memset(filter_text, '*', filter_len);
+        filter_text[filter_len] = '\0';
+    }
+
+    draw_line(0, 0, "%*s%s", title_len, "", (filter_text + doffset));
```
Cloudef, line 231: Ditto here

lib/renderers/cairo.h
robertgzr: hmm I pushed an update doing it like that. It's nicer for simplicity but feels/looks less nice I think... I'll leave it up to you

lib/renderers/curses/curses.c
```diff
@@ -221,7 +221,13 @@ render(const struct bm_menu *menu)
         doffset -= (prev ? prev : 1);
     }
 
-    draw_line(0, 0, "%*s%s", title_len, "", (menu->filter ? menu->filter + doffset : ""));
+    char *filter_text = (menu->filter ? menu->filter : "");
```
Cloudef, line 224: should be `menu->filter ? menu->filter + doffset : "")` instead, adding + doffset later may cause invalid read.

lib/renderers/curses/curses.c
```diff
@@ -221,7 +221,13 @@ render(const struct bm_menu *menu)
...
+    size_t filter_len = strlen(filter_text);
+    if (menu->password) {
+        draw_line(0, 0, "%*s%s", title_len, "");
+    } else {
+        draw_line(0, 0, "%*s%s", title_len, "", (filter_text + doffset));
```
Cloudef, line 229: Then this can be just `draw_line(0, 0, "%*s%s", title_len, "", filter_text);`

lib/renderers/cairo.h
```diff
@@ -247,7 +247,15 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
     paint.cursor = menu->cursor;
     paint.pos = (struct pos){ (menu->title ? 2 : 0) + result.x_advance, vpadding };
     paint.box = (struct box){ (menu->title ? 2 : 4), 0, vpadding, vpadding, width - paint.pos.x, ascii_height };
-    bm_cairo_draw_line(cairo, &paint, &result, "%s", (menu->filter ? menu->filter : ""));
+
+    char *filter_text = (menu->filter ? menu->filter : "");
+    size_t filter_len = strlen(filter_text);
```
Cloudef, line 252: `filter_len` is not used for anything? Can be removed.

lib/renderers/cairo.h
```diff
@@ -247,7 +247,15 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
     paint.cursor = menu->cursor;
     paint.pos = (struct pos){ (menu->title ? 2 : 0) + result.x_advance, vpadding };
     paint.box = (struct box){ (menu->title ? 2 : 4), 0, vpadding, vpadding, width - paint.pos.x, ascii_height };
-    bm_cairo_draw_line(cairo, &paint, &result, "%s", (menu->filter ? menu->filter : ""));
+
+    char *filter_text = (menu->filter ? menu->filter : "");
```
Cloudef, line 251: Can be `const char *filter_text = ...` since not modified.

lib/renderers/curses/curses.c
Cloudef: Also can be `const char *filter_text = ...` since not modified.

lib/renderers/curses/curses.c
```diff
@@ -221,7 +221,13 @@ render(const struct bm_menu *menu)
         doffset -= (prev ? prev : 1);
     }
 
-    draw_line(0, 0, "%*s%s", title_len, "", (menu->filter ? menu->filter + doffset : ""));
+    char *filter_text = (menu->filter ? menu->filter : "");
+    size_t filter_len = strlen(filter_text);
```
Cloudef, line 225: Ditto here, this is now unused.

Cloudef:
Looks good, thanks.

PR #160, opened 2021-03-29
Add 'Usage' section to man page with keybindings

### LINKED ISSUE #159: Adding keybindings to man page
The keybindings aren't listed anywhere in the manual pages. I'd like to add them, but I want to know where you think is the best place.

My idea is to add a section beneath **Description** titled **Usage**, then add the table of keybindings in that section. I've implemented this in a6af04c. If that looks good, I'll create a PR for it.

kdkasad:
Adds keybindings table found in README to the `bemenu(1)` manual page. Keybindings are listed under a new **Usage** section which comes after the **Description** section.

Closes #159

### DIFF
--- a/man/bemenu.1
+++ b/man/bemenu.1
@@ -41,6 +41,55 @@ is a special-case invocation of
 where the input is a list of executables in the $PATH directories,
 and the selection gets executed.
 
+.SH USAGE
+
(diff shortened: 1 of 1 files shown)


Cloudef:
Thanks

PR #162, opened 2021-04-03
client: use getline(3)

nmeum:
Not sure why the code didn't use this function previously, but
getline(3) is mandatory since POSIX.1-2008 and should thus be supported
by all maintstream operating systems. Glibc requires a corresponding
_POSIX_C_SOURCE feature test macro to be set, which has been added to
the GNUmakefile.

The changes proposed here make the code easier to read and should also
make it more efficient memory-wise as it does not necessarily cause
the entire input to be read into memory once.

---

Haven't tested this extensively yet.

### DIFF
--- a/client/bemenu.c
+++ b/client/bemenu.c
@@ -14,48 +14,27 @@ read_items_to_menu_from_stdin(struct bm_menu *menu)
 {
     assert(menu);
 
-    size_t step = 1024, allocated;
-    char *buffer;
+    ssize_t n;
+    size_t llen = 0;
+    char *line = NULL;
 
-    if (!(buffer = malloc((allocated = step))))
-        return;
-
-    // TODO: do not read everything into memory
-    size_t read;
-    while ((read = fread(buffer + (allocated - step), 1, step, stdin)) == step) {
-        void *tmp;
-        if (!(tmp = realloc(buffer, (allocated += step)))) {
-            free(buffer);
-            fprintf(stderr, "Out of memory\n");
-            return;
-        }
-        buffer = tmp;
-    }
-
-    // make sure we can null terminate the input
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -12,7 +12,7 @@ WARNINGS = -Wall -Wextra -Wpedantic -Wformat=2 -Wstrict-aliasing=3 -Wstrict-over
 	-Wfloat-equal -Wcast-align -Wpointer-arith -Wchar-subscripts -Warray-bounds=2 -Wno-unknown-warning-option
 
 override CFLAGS ?= -g -O2 $(WARNINGS)
-override CFLAGS += -std=c99
+override CFLAGS += -std=c99 -D_POSIX_C_SOURCE=200809L
```
Cloudef, line 15: this should go to CPPFLAGS instead, also does this conflict with -D_DEFAULT_SOURCE?

nmeum: Oh, I didn't see that you already have `_DEFAULT_SOURCE` enabled. The [documentation for `_DEFAULT_SOURCE`](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html#index-_005fDEFAULT_005fSOURCE) states the following:

> If you define this macro, most features are included apart from X/Open, LFS and GNU extensions: the effect is to enable features from the 2008 edition of POSIX, as well as certain BSD and SVID features without a separate feature test macro to control them. 

Since this macro already enables the POSIX.1-2008 features, we don't need to set `_POSIX_C_SOURCE`. I don't actually use glibc, so please check on your system, but I did some quick tests on Debian which confirm this.

Removed the Makefile modifications and adjusted the commit message.

Cloudef:
Looks good, thanks

PR #167, opened 2021-04-25
x11: Align -m argument interpretation with dmenu

### LINKED ISSUE #164: Behaviour of -m is incompatible with dmenu
While switching my dwm setup from dmenu to bemenu, I noticed that bemenu's current `-m` behaviour is incompatible with dmenu's original `-m` behaviour. With dmenu, monitor indices start at 0 and a value of -1 (the default) is used to spawn dmenu on the current monitor. With bemenu, monitor indices start 1 and a value of 0 (the default) is used to spawn on the current monitor. Would you be interested in align the default -m behaviour with the dmenu behaviour? I only tested this with the x11 backend, not with the wayland backend.

nmeum:
With dmenu, monitor indices start at 0 and a value of -1 (the default)
is used to spawn dmenu on the current monitor. While bemenu strives to
be compatible with dmenu, bemenu monitor indices previously started at 1
and a value of 0 (the default) was used to spawn on the current monitor.

This commit aligns the behaviour of bemenu's x11 backend with dmenu. For
this purposes, the affected code in the x11 backend is synced with the
current dmenu implementation. While doing so the monitor type has also
been switched from a uint32_t to a int32_t.

Fixes #164 

---

TODO:

* [ ] Align behaviour of the wayland backend as well? Presently, the wayland backend uses a value of -1 to display the menu on all monitors while the x11 backend uses -1 to display the menu on the active monitor. Not sure how that should be resolved.
* [x] Change the default monitor value to -1? Presently, 0 is still used (-1 would be a better default, at least for the x11 backend).
* [x] Update documentation in man page.

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -278,6 +278,9 @@ do_getopt(struct client *client, int *argc, char **argv[])
      * Either break the interface and make them --sf, --sb (like they are now),
      * or parse them before running getopt.. */
 
+    /* Set default monitor index to -1 (maybe be overwritten with -m below). */
+    client->monitor = -1;
+
     for (optind = 0;;) {
         int32_t opt;
 
--- a/lib/bemenu.h
+++ b/lib/bemenu.h
@@ -502,13 +502,13 @@ BM_PUBLIC bool bm_menu_get_bottom(struct bm_menu *menu);
 
 /**
  * Display menu at monitor index.
- * Indices start from 1, pass 0 for active monitor (default).
+ * Indices start at 0, a value of -1 can be passed for the active monitor (default).
  * If index is more than amount of monitors, the monitor with highest index will be selected.
  *
  * @param menu bm_menu instance to set monitor for.
- * @param monitor Monitor index starting from 1.
+ * @param monitor Monitor index starting from 0, or -1 for the active monitor.
  */
-BM_PUBLIC void bm_menu_set_monitor(struct bm_menu *menu, uint32_t monitor);
+BM_PUBLIC void bm_menu_set_monitor(struct bm_menu *menu, int32_t monitor);
 
 /**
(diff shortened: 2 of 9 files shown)


Cloudef:
On wayland you can't rely on the indices, so on wayland the best way to choose monitor is by monitor name. It could be possible to generate indices for wayland from output's physical locations. -1 as default makes sense.

nmeum:
> On wayland you can't rely on the indices, so on wayland the best way to choose monitor is by monitor name.

So basically, with Wayland the argument to `-m` should not be an integer but a string specifying a specific monitor name and selection by index is not really supported by the Wayland backend? If so: That should probably be mentioned in the man page and I think the following code from the Wayland backend, which increments the `monitor` variable, could be removed then, right?

https://github.com/Cloudef/bemenu/blob/e74224a40633e6a03d19f0b0844cca08792e6a31/lib/renderers/wayland/wayland.c#L288-L292

Seems to me the `monitor` variable is only written but not read in that function.

Cloudef:
The monitor variable is read in the if statement above. But yeah, that code is kind of useless for wayland. Numerical indices should be handled based on physical position instead.

nmeum:
Should I just leave the wayland code as is then? :thinking:

Cloudef:
Can be left as it is I think.

Cloudef:
I guess this is ready to be merged?

nmeum:
> I guess this is ready to be merged?

I guess so. I think the documentation for the `-m` command could be improved a bit and maybe it makes sense to document that the library function `bm_menu_set_monitor` should only be used with the X11 backend, but I guess that can also be done separately.

Cloudef:
That can be done separately. I'll try to come up with patch to do proper indexing on wayland based on the physical locations of outputs.

PR #174, opened 2021-05-21
wayland: Allow showing the menu on the focused monitor

robertgzr:
This adds an alias 'focused' for selecting the current monitor, which
becomes the default on x11 and wayland. The previous wayland default of
displaying on all outputs moves under '-2' or 'all'.

ref: https://github.com/Cloudef/bemenu/issues/102#issuecomment-604562234

Signed-off-by: Robert Günzler <r@gnzler.io>


Cloudef:
Looks good, thanks!

PR #178, opened 2021-06-06
Handle multiple seats (more) correctly

vilhalmer:
Previously, any seat without a keyboard could destroy our selected
keyboard. Now, select by seat instead and only destroy the keyboard if
it vanishes from that seat. This isn't actually multi-seat support, but
at least it will allow bemenu to accept input.


Cloudef:
Thanks

PR #179, opened 2021-06-10
Ignore hidden files in `bemenu`

Ma27:
On most systems it's fairly uncommon to have hidden files within
`$PATH`, but in case of NixOS this is fairly regularly the case as we
have the convention to add `foo` and `.foo-wrapped` to a `bin`-directory
if `foo` is supposed to be a wrapper which sets e.g. env-vars for the
program.

These `.foo-wrapped`-executables are almost never needed, but are shown
in `bemenu` right at the start, so I modified the selection code to skip
hidden files which is also what `dmenu` does for instance.


Cloudef:
Thank you

PR #197, opened 2021-08-14
fix scaling caused issue on window redimension on wayland

stacyharper:
This caused issue when using the -b (bottom) and -l (line) arguments on
scaled outputs.

When using a scaled output, the set_size use a wrong value as height.
We generated a scalled buffer so we used a scale x too high size.

We just have to divide the scaling to use a good size.


Cloudef:
Thanks

PR #200, opened 2021-08-15
mouse and touch support on wayland

### LINKED ISSUE #199: Mouse and Touchscreen support
I'm working on this at the moment (for wayland). 

I create this issue to track this subject.

stacyharper:
resolve: https://github.com/Cloudef/bemenu/issues/199

My main goal is to offer [Sxmo](https://sr.ht/~mil/sxmo/) a way to Wayland. We rely mainly on dmenu so we have to improve bemenu which is the best solution at this time.

To allow convergent usage (desktop + mobiles) I made those choices of inputs : 

```
For bar mode:    

    In pointer mode (mouse):

    - Mouse wheel select the next/previous entry
    - Click return the entry

For line mode:

    In pointer mode (mouse):

    - Moving the mouse over entries highlight them
    - Click return the entry
    - Mouse wheel scroll pages

    In touch mode (mobile);

    - Touch and move over entries highlight them
    - Release touch on an entry return this entry
    - Release touch outside of bemenu cancel the selection (left or right of bemenu)
    - Release touch above or bellow bemenu scroll the page previous or next
    - You can also double touch and scroll up or down to scroll the pages
```

One question remains to me: 

- Is the redraw could be optimized ? With this patch, I removed the `last_code` toggle cause we now have pointer events. I could not find a way to not redraw the window if my cursor moved from 5 px in the exact same entry but I think we should !

This is a big patch and I made decisive choices. I open to ideas and improvement !

This need Sway to support touch seat operation to works correctly with touches: 

https://github.com/swaywm/sway/pull/6455

To demonstrate the touch support, I recorded this for Sxmo : 

https://tube.missbanal.net/videos/watch/7a7c1bbe-ac19-4593-9e56-e46c82cbb20c

### DIFF
--- a/lib/menu.c
+++ b/lib/menu.c
@@ -293,6 +293,7 @@ uint32_t
 bm_menu_get_line_height(struct bm_menu *menu)
 {
     assert(menu);
+
     return menu->line_height;
 }
 
@@ -308,6 +309,34 @@ bm_menu_get_cursor_height(struct bm_menu *menu)
 {
     assert(menu);
     return menu->cursor_height;
+}
+
+uint32_t
+bm_menu_get_height(struct bm_menu *menu)
+{
+    assert(menu);
+
+    uint32_t height = 0;
+
+    if (menu->renderer->api.get_height) {
+        height = menu->renderer->api.get_height(menu);
+    }
+
+    return height;
+}
+
+uint32_t
+bm_menu_get_width(struct bm_menu *menu)
+{
+    assert(menu);
+
+    uint32_t width = 0;
+
+    if (menu->renderer->api.get_width) {
+        width = menu->renderer->api.get_width(menu);
+    }
+
+    return width;
 }
 
 bool
@@ -707,6 +736,41 @@ bm_menu_poll_key(struct bm_menu *menu, uint32_t *out_unicode)
     return key;
 }
 
+struct bm_pointer
+bm_menu_poll_pointer(struct bm_menu *menu)
+{
+    assert(menu);
+
+    struct bm_pointer pointer = {0};
+
+    if (menu->renderer->api.poll_pointer)
+        pointer = menu->renderer->api.poll_pointer(menu);
+
+    return pointer;
+}
+
+struct bm_touch
+bm_menu_poll_touch(struct bm_menu *menu)
+{
+    assert(menu);
+
+    struct bm_touch touch = {0};
+
+    if (menu->renderer->api.poll_touch)
+        touch = menu->renderer->api.poll_touch(menu);
+
+    return touch;
+}
+
+void
+bm_menu_release_touch(struct bm_menu *menu)
+{
+    assert(menu);
+
+    if (menu->renderer->api.release_touch)
+        menu->renderer->api.release_touch(menu);
+}
+
 static void
 menu_next(struct bm_menu *menu, uint32_t count, bool wrap)
 {
@@ -724,6 +788,41 @@ menu_prev(struct bm_menu *menu, uint32_t count, bool wrap)
         menu->index--;
     } else if (wrap) {
         menu->index = count - 1;
+    }
+}
+
+static void
+menu_point_select(struct bm_menu *menu, uint32_t posx, uint32_t posy, uint32_t displayed)
+{
+    (void) posx;
+    uint32_t selected_line = posy / (bm_menu_get_height(menu) / displayed);
+    uint16_t current_page_index = menu->index / menu->lines;
+
+    if (0 == selected_line) { // Mouse over title bar
+        return;
+    }
+
+    if (selected_line >= displayed) { // This might be useless
+        return;
+    }
+
+    menu->index = current_page_index * menu->lines + (selected_line - 1);
+}
+
+static void
+menu_scroll_down(struct bm_menu *menu, uint16_t count)
+{
+    if (menu->index / menu->lines != count / menu->lines) { // not last page
+        menu->index = ((menu->index / menu->lines) + 1) * menu->lines;
+    }
+}
+
+static void
+menu_scroll_up(struct bm_menu *menu, uint16_t count)
+{
+    (void) count;
+    if (menu->index / menu->lines) { // not first page
+        menu->index = ((menu->index / menu->lines) - 1) * menu->lines + menu->lines - 1;
     }
 }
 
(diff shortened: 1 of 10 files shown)


Eluminae:
Looks like I completly forgotten touch which looks like a different thing for Wayland.

Eluminae:
Okay I added the touch support as well. So yea, wl_pointer are different from wl_touch. I discovered all my works was useless when I disconnected my keyboard from my Pinephone.

I made some changes then: 

In pointer mode (mouse):

- Moving the mouse over entries highlight them
- Release click return the entry
- Mouse wheel scroll pages

In touch mode (mobile);

- Touch and move over entries highlight them
- Release touch return the entry

The only missing part is : I cant find a way to handle page scrolling with touches.

My touch event still occurs when I leave the window but I dont have relative position. I'd like to have some negative y that can mean "I'm above the window" and a y > height that can mean "I'm bellow the window" :(

Cloudef:
You probably can't receive those events outside of the window. What you probably can do to workaround this is to have larger window, the surrounding area can be invisible. Subsurfaces may work as well.

Eluminae:
I just added normal mode pointer support. I used mouse wheel and normal click to browse entry and select the highlighted one.

Eluminae:
> You probably can't receive those events outside of the window. What you probably can do to workaround this is to have larger window, the surrounding area can be invisible. Subsurfaces may work as well.

I thought about it. I dunno how easy it could be implemented and if other things will just works as well. I'll give it a try !

lib/renderers/wayland/wayland.c
```diff
@@ -214,6 +210,77 @@ poll_key(const struct bm_menu *menu, unsigned int *unicode)
...
+    bm_pointer.axes[0].value = event->axes[0].value;
+    bm_pointer.axes[0].discrete = event->axes[0].discrete;
+    bm_pointer.axes[1].valid = event->axes[1].valid;
+    bm_pointer.axes[1].value = event->axes[1].value;
+    bm_pointer.axes[1].discrete = event->axes[1].discrete;
+    bm_pointer.axis_source = event->axis_source;
+
+    bm_pointer.button = BM_POINTER_KEY_NONE;
+    switch (event->button) {
+        case 272:
```
Eluminae, line 237: I should find an enum for this

Eluminae:
Another issue I got is: If there is a window behind bemenu. I will continue to receive motion event with x and y relative to the other window. With all those constraint, I'm not sure I can reproduce the sxmo dmenu gesture behaviors.

Eluminae:
I also though about adding pagination entries above and behind the others. Those entries could be toggelable with a new argument and we could rely on this in sxmo.

Cloudef:
> Another issue I got is: If there is a window behind bemenu. I will continue to receive motion event with x and y relative to the other window. With all those constraint, I'm not sure I can reproduce the sxmo dmenu gesture behaviors.

Maybe better approach would be to have compositor do touch -> pointer event translation sort of like sxmo seems to do under X11 with the gesture lib.

Eluminae:
> Maybe better approach would be to have compositor do touch -> pointer event translation sort of like sxmo seems to do under X11 with the gesture lib.

Could you detail this please ? I am a little bit slow

Eluminae:
I think I got it. You mean keep old pos on each event and building a relative to the last pos vector and aggregate it to a full vector.

It make sense ! And it will eventually allow to determine if we leaved the window from a starting position ! I'm on it

Eluminae:
I tried something using relative motion vectors but now I got an issue when there is no window behind. Leaving the window and entering it again by another point break the vector consistency

I'm not sure how to handle it :(

Cloudef:
> I think I got it. You mean keep old pos on each event and building a relative to the last pos vector and aggregate it to a full vector.
> 
> It make sense ! And it will eventually allow to determine if we leaved the window from a starting position ! I'm on it

I meant in X11 sxmo seems to use daemon or window manager (I did not look that deep) to translate touch / gestures into normal mouse events. This could also be done in compositor like sway, thus bemenu not needing touch handling at all. Compositor doesn't have the problem with needing larger window to capture surrounding events.

Eluminae:
To be honest:

I dont think it is doable or desirable to handle touch page scrolling on Wayland with the solution we thought about.

So for the moment I made the pos_x and pos_y fields of the bm_touch_point as integer and not unsigned integer. This allow the rendered to give negative value and valuer higher than the window box sizes, relatively to the top left position as 0, 0. This would eventually allow this behavior to be implemented in the future.

I'll try to handle this issue with a completely different patches. I think that my idea of additional paginations entries is better !

I unWIP this so we can start review it, improve it, and finally merge it.

Eluminae:
> I meant in X11 sxmo seems to use daemon or window manager (I did not look that deep) to translate touch / gestures into normal mouse events. This could also be done in compositor like sway, thus bemenu not needing touch handling at all. Compositor doesn't have the problem with needing larger window to capture surrounding events.

I'm not really sure how dmenu does it but it surely doesnt got the same
constraints as with Wayland.

Translating touch to pointer would not change that much. We cant have
pointer position outside of the bemenu surface anyway.

And if I understand your statement, you mean configuring the input
touchscreen as a classic mouse. Maybe it is possible but it would be
very sad to loose touch support for all the compositor just for bemenu
:(

Eluminae:
Okay I just implemented two finger scroll. It works nicely !

Cloudef:
Can you fix the build failures for the osx? https://github.com/Cloudef/bemenu/pull/200/checks?check_run_id=3363969487
It seems to automatically do static analysis there which is nice.

Eluminae:
I fixed osx checks in a dedicated commit

Eluminae:
I played with wev a little bit. I discovered that 

with pointer :

Holding click and leave the window borders continue to trigger event motion relatively to the initial surface coordinates

with touch : 

Touching and leave the window borders continue to trigger event motion relatively to the underlying surface coordinates.

So then, it may just be an issue with Sway. I'm trying to discuss and/or open an issue with them

Eluminae:
Okay this MR need Sway to implement correctly seatop_touch : https://github.com/swaywm/sway/pull/6455

This will works without but will react crazily if there is other window arround. Plus you cannot cancel your touch releasing it outside of bemenu.

So I added what we call the "Tetrakist" patch (the code on dmenu was from him). So releasing above or bellow bemenu will scroll the page.

Eluminae:
I fixed up my commits in one. I think it is ready for review. 

I still would like some kind of notification of what bemenu will be doing if I release touch above/bellow the menu. In dmenu, the last or first entry display "Page Down" or "Page Up" or also "Last Page" or "First Page". I will probably do this in a second time. This will also allow this gesture behavior to go First or Last page, if your touch is way bellow/above the menu.

stacyharper:
I used my both pointer and touch event_masks to trigger bm_wl_window_schedule_render smartly. Bemenu no longer redraw every frame if no keybind nor pointer nor touch action happen

stacyharper:
Hello there !

Can I have some feedback about this ? :D

Cloudef:
Sorry, I'm currently super busy with work. This PR is more involved so it needs more of my review time.

stacyharper:
Np take your time, no need to hurry this. I just needed to hear that
your okay and that you will check it when more available :)

stacyharper:
I rebased and pushed a last commit to optimize redrawing when nothing actually changed in the menu. I still have to try it over time and see how it behave.

Cloudef:
The code looks very good to me. Can this be rebased against latest develop and fix the build errors:
```
 lib/renderers/wayland/wayland.c: In function ‘render’:
lib/renderers/wayland/wayland.c:47:27: error: passing argument 1 of ‘bm_menu_set_dirty’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
   47 |         bm_menu_set_dirty(menu, false);
      |                           ^~~~
In file included from lib/internal.h:4,
                 from lib/renderers/wayland/wayland.c:1:
lib/bemenu.h:786:50: note: expected ‘struct bm_menu *’ but argument is of type ‘const struct bm_menu *’
  786 | BM_PUBLIC bool bm_menu_set_dirty(struct bm_menu *menu, bool dirty);
      |                                  ~~~~~~~~~~~~~~~~^~~~
```

lib/menu.c
```diff
@@ -707,24 +749,128 @@ bm_menu_poll_key(struct bm_menu *menu, uint32_t *out_unicode)
...
-        menu->index = count - 1;
+        bm_menu_set_highlighted_index(menu, count - 1);
+    }
+}
+
+static void
+menu_point_select(struct bm_menu *menu, uint32_t posx, uint32_t posy, uint32_t displayed)
+{
+    (void) posx;
+    uint32_t selected_line = posy / (bm_menu_get_height(menu) / displayed);
```
Cloudef, line 811: this looks like division by zero can happen

lib/menu.c
```diff
@@ -707,24 +749,128 @@ bm_menu_poll_key(struct bm_menu *menu, uint32_t *out_unicode)
...
+    uint16_t current_page_index = menu->index / menu->lines;
```
Cloudef, line 812: ditto here

lib/menu.c
```diff
@@ -707,24 +749,128 @@ bm_menu_poll_key(struct bm_menu *menu, uint32_t *out_unicode)
...
+        return;
     }
+
+    bm_menu_set_highlighted_index(menu, current_page_index * menu->lines + (selected_line - 1));
+}
+
+static void
+menu_scroll_down(struct bm_menu *menu, uint16_t count)
+{
+    if (menu->index / menu->lines != count / menu->lines) { // not last page
```
Cloudef, line 828: ditto here, if `menu->lines == 0` is an impossible condition, assert should be used to indicate that.

lib/menu.c
```diff
@@ -707,24 +749,128 @@ bm_menu_poll_key(struct bm_menu *menu, uint32_t *out_unicode)
...
+        bm_menu_set_highlighted_index(menu, ((menu->index / menu->lines) + 1) * menu->lines);
+    }
+}
+
+static void
+menu_scroll_up(struct bm_menu *menu, uint16_t count)
+{
+    (void) count;
+    if (menu->index / menu->lines) { // not first page
```
Cloudef, line 837: ditto here

lib/menu.c
```diff
@@ -707,24 +749,128 @@ bm_menu_poll_key(struct bm_menu *menu, uint32_t *out_unicode)
...
+        bm_menu_set_highlighted_index(menu, ((menu->index / menu->lines) - 1) * menu->lines + menu->lines - 1);
```
Cloudef, line 838: and here :)

lib/menu.c
```diff
@@ -707,24 +749,128 @@ bm_menu_poll_key(struct bm_menu *menu, uint32_t *out_unicode)
...
     }
+
+    bm_menu_set_highlighted_index(menu, current_page_index * menu->lines + (selected_line - 1));
+}
+
+static void
+menu_scroll_down(struct bm_menu *menu, uint16_t count)
+{
+    if (menu->index / menu->lines != count / menu->lines) { // not last page
+        bm_menu_set_highlighted_index(menu, ((menu->index / menu->lines) + 1) * menu->lines);
```
Cloudef, line 829: here

lib/menu.c
```diff
@@ -954,4 +1103,207 @@ bm_menu_run_with_key(struct bm_menu *menu, enum bm_key key, uint32_t unicode)
...
+    for (size_t i = 0; i < nmemb; ++i) {
+        struct bm_touch_point point = touch.points[i];
+        if (point.event_mask & TOUCH_EVENT_DOWN) {
+            count_down += 1;
+        }
+    }
+
+    if (count_down == 2) {
+        int16_t scroll_count = 0;
+        int16_t scroll_directions[nmemb];
```
Cloudef, line 1200: can we make this not use VLA? I assume `nmemb` won't ever be more than 5 so we can hardcode the scroll_directions to [5]?

stacyharper:
> The code looks very good to me. Can this be rebased against latest develop and fix the build errors:
> 
> ```
>  lib/renderers/wayland/wayland.c: In function ‘render’:
> lib/renderers/wayland/wayland.c:47:27: error: passing argument 1 of ‘bm_menu_set_dirty’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
>    47 |         bm_menu_set_dirty(menu, false);
>       |                           ^~~~
> In file included from lib/internal.h:4,
>                  from lib/renderers/wayland/wayland.c:1:
> lib/bemenu.h:786:50: note: expected ‘struct bm_menu *’ but argument is of type ‘const struct bm_menu *’
>   786 | BM_PUBLIC bool bm_menu_set_dirty(struct bm_menu *menu, bool dirty);
>       |                                  ~~~~~~~~~~~~~~~~^~~~
> ```

I think I will need help on this point. To make re-drawing more efficient I added this "dirty" boolean. But then the only way I found to clear it is to do it in the backend render function. But that break the const type. So imho my design on this is wrong and maybe there is better ways. 

btw this re-drawing isnt perfect and sometime the menu show itself in an inconsistent way. Like if I missed a necessary redraw. Probably cause of double buffering somehow

lib/menu.c
stacyharper: should be fixed !

lib/menu.c
stacyharper: should be fixed !

lib/menu.c
stacyharper: should be fixed !

lib/menu.c
stacyharper: should be fixed !

lib/menu.c
stacyharper: should be fixed !

Cloudef:
I don't think the dirty bit needs to be part of a public api, it can be simply member of the menu that the library uses internally. The only time you call it from outside the library is to set the menu "dirty" during the start, the library probably internally should set itself dirty when items are added/removed/selection changes/etc. I don't have a problem with dropping the const in the function signature.

>btw this re-drawing isnt perfect and sometime the menu show itself in an inconsistent way. Like if I missed a necessary redraw

Does this happen under X11 as well?

stacyharper:
> I don't think the dirty bit needs to be part of a public api, it can be simply member of the menu that the library uses internally. The only time you call it from outside the library is to set the menu "dirty" during the start, the library probably internally should set itself dirty when items are added/removed/selection changes/etc. I don't have a problem with dropping the const in the function signature.

I see, I'll work on this later today !

> >btw this re-drawing isnt perfect and sometime the menu show itself in an inconsistent way. Like if I missed a necessary redraw
> 
> Does this happen under X11 as well?

Nah only on wayland cause we dont use the dirty bit in other backends
yet. They will need it when we'll write the touch and pointer parts of
them. Cause they only react to keyboard, their redrawing is still
simple.

lib/menu.c
stacyharper: I think this code came from a long time. Now I enforce 2 in the backend wayland so lets just enforce 2 everywhere. It would be easy to change anyway.

Cloudef:
Need to fix the flatpak job..

stacyharper:
> Need to fix the flatpak job..

Are you on this ? Or should I do something ? I dont know much about github CI but I think we only miss a dependency

Cloudef:
I'm fixing it right now, I'll let you know when to rebase

Cloudef:
@t-8ch That was oversight on my part. I don't see a reason why that function should be removed, I restored it here: https://github.com/Cloudef/bemenu/pull/246 and also made the other run functions public.

t-8ch:
@Cloudef Thanks!

Cloudef:
@t-8ch merged and made 0.6.6 to fix that warning for you, also since you could use the function symbols weren't hidden properly :)

t-8ch:
Thanks! I have switched to `bm_menu_run_with_events()` now, to take advantage of the new functionality, but that change is still good.

Cloudef:
Yeah with the change people your old versions won't break.

PR #201, opened 2021-08-24
Add option to configure cursor height

benjamb:
If set to 0 (the default), the height of the cursor is set to the
height of the line (as is the current behaviour).


benjamb:
I've not actually tested this on X, but I've not written anything display server specific.

Issue #128, opened 2020-08-15
Height is 1 px less


hacktivista:
Height is 1 px less. Running bemenu with -H 28 on sway renders 27px

`bemenu-run -i -m all -p '' -H 28 --fn 'monospace 12'`

![image](https://user-images.githubusercontent.com/5282749/90312187-60fd5380-ded0-11ea-9092-5cf3f8a1696e.png)

soyuz2012:
I had the same issue. I fixed this by using other fonts (works fine for Terminus)

reidevries:
I have a similar, but slightly different issue where `--line-height=22` had a height of 21 pixels, and `--line-height=23` had a height of 23 pixels. Tried it with the default font as well as fira sans, and with different font sizes, and all have the same result. I don't believe I have any DPI scaling turned on.

I am using this command to launch bemenu in sway:
`bindcode 94 exec `bemenu-run -i "$@" \
         -m "$(swayfocused)"\
         -p "" \
         --line-height=22 \
         --fn "font pango:Fira Sans Bold 10" \
         --tb "#000000" \
         --tf "#f7f7f7" \
         --fb "#000000" \
         --ff "#f7f7f7" \
         --nb "#000000" \
         --nf "#f7f7f7" \
         --hb "#f7f7f7" \
         --hf "#000000" \
         --sb "#000000" \
         --sf "#979797" `

![The result using `--line-height=22` which has a pixel height of 21](https://user-images.githubusercontent.com/25862413/135583393-ee1b8a60-f264-473d-944b-8b7ca0edcf8f.png)
![The result using `--line-height=23` which has a pixel height of 23](https://user-images.githubusercontent.com/25862413/135583399-fa071df4-b18a-42c0-b97b-c966c32be638.png)

PR #207, opened 2021-10-01
Fix incorrect line height 

reidevries:
Hi, I found a fix for the [incorrect line height rendering issue](https://github.com/Cloudef/bemenu/issues/128).

It seems that `vpadding` for the text is calculated with a integer division by 2, with the same padding applied above and below the text, which effectively quantizes the overall height to odd numbers. The solution I used was to calculate a separate `vpadding` for the top and bottom of the text, so when an even number is given for the line height, `vpadding_t = vpadding_b+1`. Now testing across `--line-height=23`, `--line-height=24` and `--line-height=28` all result in the correct line height being rendered.

I'm not really that familiar with cairo or the rest of the project so pls forgive me if the fix seems hacky or if there is a better solution.

### DIFF
--- a/lib/renderers/cairo_renderer.h
+++ b/lib/renderers/cairo_renderer.h
@@ -237,15 +237,16 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
     ascii_height = result.height;
     paint.baseline = result.baseline;
 
-    int32_t vpadding = height == 0 ? 2 : (height - ascii_height) / 2;
+    int32_t vpadding_t = height == 0 ? 2 : (height - ascii_height) / 2;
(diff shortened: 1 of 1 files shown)


Cloudef:
I think the real problem here is that it's using ascii_height instead of height for drawing. (Originally introduced in this commit: https://github.com/Cloudef/bemenu/commit/d3c43973c8fe04e625b1ae62696954d3abb65699) The same issue probably can be fixed by assigning height to ascii_height if ascii_height is less than height. I guess the ascii_height was introduced when font is actually larger (in pixels) than the specified font size to avoid clipping.

reidevries:
I amended it using ur suggestion, does the code look ok now?

Tested again and it renders the correct pixel height

Cloudef:
Thanks

Issue #194, opened 2021-08-03
Feature request: Horizontal customizability (menu width, horizontal alignment, text alignment)


Hubro:
bemenu seems to always cover 100% of the width of the monitor. I have several very wide monitors, so bemenu is taking up a ton of space.

![image](https://user-images.githubusercontent.com/597206/128013803-cb59cec8-3a9a-417b-af38-8d4de65e317e.png)

(This is just a 16:9 monitor, at home I have both a 21:9 and 32:9 where this issue is much more pronounced)

It would be super neat to have some more horizontal customizability, like:

- Menu width (in `px` or `%`)
- Horizontal menu position (left, center, right)
- Text alignment (left, center, right)

It would look so much nicer if I could limit the width of bemenu to (for example) around 400px:

![image](https://user-images.githubusercontent.com/597206/128013882-9f523b25-2db3-4714-a45e-1988b842b74b.png)

On my 32:9 monitor it would also be a massive user experience boost to be able to center the menu. Currently the menu covers the entire width of the monitor and the text is at the far left. On a monitor that wide, it means I have to turn my head to see the text, plus the menu covers the bottom half of the monitor, which is a 5120x720px area.

Cloudef:
Good idea, this should not be that hard I think.

stacyharper:
Center mode has been merged recently -c. I'm also working on disabling the spacing with somekind of (-s).

stacyharper:
One alternative I can think about is keeping empty lines when on the
last page to keep the same height

stacyharper:
> I'm not seeing the `--center` option in `man bemenu`. Also, this option appears to have a set width that covers maybe 80% of the screen, not sure if this is expected.

Maybe because you openned /usr/man/… and not /usr/local/man/… where
your builded version has been installed

Hubro:
Currently, both `-c/--center` and `-b/--bottom` are more or less unusable. Since it's not possible to set a static height, the input prompt keeps jumping around as the number of search results change.

For `-c/--center` to be usable, there also *needs* to be a `--height` parameter to lock the result box to a certain height.

For `-b/--bottom` to be usable, there could also be an option to display the prompt *below* the box, rather than on top. In that case, the prompt would not move as the height of the box changed.

These would both be great additions, but for me, it's much more critical to be able to limit with *width* of the bemenu box. Here's how it currently behaves on my 5120x1440 monitor:

[...]

PR #209, opened 2021-10-17
Add horizontal alignment and improve center alignment

llunacb:
Adds a`--margin` option which can be used to set the empty space on the left and right of the menu. Also makes the behavior of `--center` more predictable and work on both x11 and wayland. Related to the discussion in #194.


Cloudef:
Thanks, one thing I'm starting to be concerned is the pollution of single character options, we may want to keep advanced options as a long options only.

PR #211, opened 2021-11-02
lib/renderers/curses/curses.c: always use "%s"-style format for printf()-style functions

trofi:
`ncuses-6.3` added printf-style function attributes and now makes
it easier to catch cases when user input is used in palce of format
string when built with CFLAGS=-Werror=format-security:

    lib/renderers/curses/curses.c:234:9:
      error: format not a string literal and no format arguments [-Werror=format-security]
      234 |         mvprintw(0, 0, menu->title);
          |         ^~~~~~~~

Let's wrap all the missing places with "%s" format.


Cloudef:
Thanks!

PR #214, opened 2021-11-23
Instruction for installing dependencies

jjtt:
I had to hunt for the specific packages needed to be able to build on my fresh Ubuntu install.

I think having an example like this in the README.md would have helped me.


Cloudef:
Thanks

PR #220, opened 2021-12-26
Add relative width option

begss:
It works on Wayland and X11 and acts as a complement to margin. Exact
behavior is as follows:
- If width factor is 0, width minus margin is used.
- If width multiplied by factor is greater than width minus margin,
  width minus is used. (so margin may be used to make sure that bemenu
  is at least N pixels away from the view border)
- Otherwise width multiplied by factor is used.

Also, what do you think about using [scdoc](https://git.sr.ht/~sircmpwn/scdoc) to generate manpage instead of plain roff? If it's fine, than I'll make a follow-up PR with necessary changes plus `--margin` and `--width-factor` documented.


Cloudef:
>Also, what do you think about using scdoc to generate manpage instead of plain roff? If it's fine, than I'll make a follow-up PR with necessary changes plus --margin and --width-factor documented.

Yeah this would be welcome

begss:
OK, I pushed it here instead of separate PR.

Cloudef:
Can you add scdoc to the dependencies in the readme and also install it to the CI https://github.com/Cloudef/bemenu/blob/master/.github/workflows/ci.yml#L15 ?

begss:
Done.

Cloudef:
Thanks! Merged.

PR #226, opened 2022-02-09
makefile: Add uninstall target

earboxer:
Uninstalling is useful if your distro updates bemenu (and running that tries to load in your older, development-installed library rather than the newer, distro-installed library).


Cloudef:
Thanks

Issue #252, opened 2022-03-11
`--ch` cursor height option not being applied


mpahsu:
It seems the `--ch` option to change the cursor height is not being applied correctly.

**Steps to reproduce:**
1. `bemenu-run --line-height 32 --ch 16`

**Expected result:**
Cursor to be half of the height of the bemenu line.

**Actual result:**
Cursor is still the full line height.

Screenshot:
![image](https://user-images.githubusercontent.com/3999823/157833372-42735c3d-bdaa-4124-876a-42e6ae3c3b71.png)

mpahsu:
I have raised #253 which I think fixes this.

TripleK2004:
> 1. bemenu-run --line-height 32 --ch 16

This hasn't yet been solved in 0.6.7 ? There is no change in height upon running the above command.
Installed bemenu from community repository, arch linux. `bemenu-wayland` pkg

mpahsu:
> This hasn't yet been solved in 0.6.7 ?

Yeah, I think v0.6.7 was cut on 21 Feb 2022 and this change was merged on 11 Mar 2022. v0.6.7 is the latest release.

@Cloudef: do you know when the next release will be cut for us to get this fix out?

PR #253, opened 2022-03-11
Fix `--ch` cursor height option parsing

mpahsu:
It seems the `--ch` option to set the cursor height is no longer working due to our reuse of the option value `0x118`.
This looks to have been introduced in 9b8da124675340c67ce28c488b63da5bdcb45c7c by accident.

This change corrects its value to `0x120` which is expected further down for setting the cursor height:

https://github.com/Cloudef/bemenu/blob/c3a662d0961a696c9739b089bcdfe92cf6a89bfd/client/common/common.c#L380-L382

Fixes: #252

**Testing**

Manually ran:
`LD_LIBRARY_PATH=. BEMENU_RENDERERS=. ./bemenu-run --line-height 32 --ch 16`

Before:
image

After:
image


Cloudef:
Thanks merged

Issue #193, opened 2021-07-27
Draw borders?


Barbaross93:
I was wondering if there were plans to be able to draw borders around the bemenu window? If not, where in the code should I look if I wanted to send a PR?

Cloudef:
All the drawing stuff for x11 and wayland is in this file: https://github.com/Cloudef/bemenu/blob/master/lib/renderers/cairo_renderer.h

Barbarossa93:
Shoot, this might be a bit more than I can handle :(

Is there always a one pixel border drawn? If I can find the appropriate line(s) that does this, it may be as simple as adding a border width to it.

Cloudef:
> Is there always a one pixel border drawn? If I can find the appropriate line(s) that does this, it may be as simple as adding a border width to it.

Do you mean something like x11 window border? Wayland doesn't have such concept, so it would be more portable if you did the border drawing in the cairo_renderer.h

Barbarossa93:
No, I meant through cairo. It might just be that my line height isn't as high as it should be, but it seemed like there was a one pixel border. It's noticeable if you have a background color set for one element such as the prompt string compared to the default background.

If I'm going to do a PR, I definitely want to do it right and do it through cairo.

If I'm reading the `cairo_renderer` code correctly, it looks like the `bm_cairo_draw_line` function is the one I want to mess with? or do I want to play with `paint.box`/`paint.pos`?

Cloudef:
I think you want to use the cairo functions directly instead, the bm_cairo functions mainly work with text. Line here meaning a line of text.

Barbarossa93:
Nevermind, regarding colors, I think I figured that out :) Unfortunately things do get covered up if the border is too thick, so still not sure how to deal with this.

PR #265, opened 2022-05-31
Add option to define a border and border color

Barbaross93:
Well, I finally have something working. I suspect I will need to polish it up a bit as I wasn't sure what I was doing in terms of adding the new options. One thing to note is that the border is drawn on the inside of the cairo surface, meaning it will overlap text. Ideally, I think it makes sense to expand the cairo surface to allow room for the border, but I am unsure of how to do this.

Closes https://github.com/Cloudef/bemenu/issues/193.

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -202,6 +202,7 @@ usage(FILE *out, const char *name)
           " -H, --line-height     defines the height to make each menu line (0 = default height). (wx)\n"
           " -M, --margin          defines the empty space on either side of the menu. (wx)\n"
           " -W, --width-factor    defines the relative width factor of the menu (from 0 to 1). (wx)\n"
+          " -B, --border          defines the width of the border in pixels around the menu. (wx)\n"
           " --ch                  defines the height of the cursor (0 = scales with line height). (wx)\n"
           " --cw                  defines the width of the cursor. (wx)\n"
           " --hp                  defines the horizontal padding for the entries in single line mode. (wx)\n"
@@ -221,7 +222,8 @@ usage(FILE *out, const char *name)
           " --ab                  defines the alternating background color. (wx)\n"
           " --af                  defines the alternating foreground color. (wx)\n"
           " --scb                 defines the scrollbar background color. (wx)\n"
-          " --scf                 defines the scrollbar foreground color. (wx)\n", out);
+          " --scf                 defines the scrollbar foreground color. (wx)\n"
+          " --bdr                 defines the border color. (wx)\n", out);
 
     exit((out == stderr ? EXIT_FAILURE : EXIT_SUCCESS));
 }
@@ -274,6 +276,7 @@ do_getopt(struct client *client, int *argc, char **argv[])
         { "line-height",  required_argument, 0, 'H' },
         { "margin",       required_argument, 0, 'M' },
         { "width-factor", required_argument, 0, 'W' },
+        { "border",       required_argument, 0, 'B' },
         { "ch",           required_argument, 0, 0x120 },
         { "cw",           required_argument, 0, 0x125 },
         { "hp",           required_argument, 0, 0x122 },
@@ -294,6 +297,7 @@ do_getopt(struct client *client, int *argc, char **argv[])
         { "af",           required_argument, 0, 0x124 },
         { "scb",          required_argument, 0, 0x114 },
         { "scf",          required_argument, 0, 0x115 },
+        { "bdr",          required_argument, 0, 0x121 },
 
         { "disco",       no_argument,       0, 0x116 },
         { 0, 0, 0, 0 }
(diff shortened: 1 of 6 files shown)


Cloudef:
Yes, please redo the PR without the auto-formatting changes if possible.

Barbaross93:
Alright, I was able to clean up the mess. I mentioned earlier that the borders are drawn on the inside and that a possible solution may be to expand the cairo surface. Perhaps another strategy would be to make the width/height corrections to all the other drawn elements. What do you think?

Cloudef:
It might be better to have the border also act as a padding for the content. Expanding the canvas means extra calculation to match the desired target width / height.

Barbaross93:
So in order have the border act as padding without expanding the canvas, I will have to make the corrections to all the other drawn elements, right?

Cloudef:
Yes

Barbaross93:
Do have some clues on how to effectively go about that? Right now I'm fiddling with `paint.box` but it only seems to get me so far.

Cloudef:
You want to subtract `border_size * 2` from width and height around here https://github.com/Cloudef/bemenu/blob/1f7c009f3f6116910374490ba9247c017a2d3669/lib/renderers/cairo_renderer.h#L275 and then add `border_size` offset to every `paint.box`, may need some other small tuning but this should get you started. You can then also simplify your border drawing, by doing it here: https://github.com/Cloudef/bemenu/blob/1f7c009f3f6116910374490ba9247c017a2d3669/lib/renderers/cairo_renderer.h#L277
1. draw 0, 0, width, height rectangle with the border color
2. draw border_size, border_size, width - border_size * 2, height - border_size * 2 rectangle with the bg color

Barbaross93:
So I tried my best to follow your advice, but I just couldn't get it to work that way. What I have just pushed seems to work great, but admittedly I didn't test it thoroughly.

Barbaross93:
The biggest issue I'm running into is with vertical mode is active. Ideally, I would subtract from the maximum height and scrunch the list within the box. This means that if line height was specified, it will have to be altered from what the user specified. Alternatively, if the border is  large enough, we could reduce the number of entries show but again, this will be altered from what the user specified. In order to respect the line height and the number of lines to be shown, the borderr MUST be drawn along the outside of the cairo surface.

Cloudef:
I think adding to the window width and height would make sense. `width = min(width + border_size, monitor_width);` (and same for height as well)

Barbaross93:
Where do I alter the window width/heights? In `cairo_renderer.h` still?

Cloudef:
https://github.com/Cloudef/bemenu/blob/master/lib/renderers/x11/window.c#L67
https://github.com/Cloudef/bemenu/blob/master/lib/renderers/wayland/window.c#L306

Height is determined by `result.height` https://github.com/Cloudef/bemenu/blob/master/lib/renderers/cairo_renderer.h#L458

Barbaross93:
Alright, I am _very_ close. Ultimately when I ended up doing is to only expand the height of the window instead of expanding both the width and height. The reason I did this is because the expansion of the width would violate any set margins or width factor made by the user. Also, we can afford to cut content lengthwise as it'll get cut anyway if the user specifies a large enough margin/width factor. 

At the moment, vertical mode works perfectly. The only remaining issue involves single line mode. The for loop that creates the horizontal list of entries continues until `cl` is larger than the width. This is problematic because the loop will continue one last time until the entry list is longer than the width. The last entry, depending on how long it is, will overlap the borders. I have commented out an if statement near the bottom: It's goal is to determine how long the next entry is in pixels, add it to the current `cl` x-offset as well as the applied horizontal padding. If it ends up being longer than the width, break the loop then and there. For some reason that I can't figure out, when I begin searching for an entry, typically after I enter the third character, bemenu crashes with a segmentation fault. I'm assume there has to be a better way to attempt the check I'm making.

What do you think of this approach? Do you know why that check is causing a segmentation fault, and if so, know of a proper solution?

Barbaross93:
Welp. I think I figured it out! Let me know if I missed something.

Barbaross93:
So, I tried to keep the border drawing done early, as in, drawn before anything else. As I mentioned, this caused some issues with horizontal mode where entries were being drawn over the border. While this works for the most part, if an entry takes up the entire width, then nothing gets shown. This is certainly not ideal. AFAIK, the only way to work around this is to draw the border at the very end. 

Thoughts?

Barbaross93:
Hey @Cloudef 

Sorry for all of the previous noise; I think I found it helpful just to write my thoughts out.

Regardless, I feel pretty confident about the current implementation. I've been using it for a while now without issues. When you get a chance, another review would be greatly appreciated.

This is the most involved PR I've made yet, so it would mean a ton to me to get this merged :)

Cloudef:
I looked it through. It looks good. Merging!

PR #267, opened 2022-06-02
Add option to specify alternating background/foreground colors for entries

Barbaross93:
This PR adds an option to specify an alternating background/foreground color for the entries. The idea for this is to make it even easier to see when an entry begins and end when using the single line mode. In vertical mode, it can look pretty nice :)


Cloudef:
Looks good, thanks

quite:
@Barbaross93  @Cloudef 

Sorry, late to the party. I think that alternative colors should have defaulted to not being different. Or, it should be optional to turn on. As the release happens now, users will suddenly see alternating colors, and will have to manually configure the colors to get rid of them.

zdykstra:
> @Barbaross93 @Cloudef
> 
> Sorry, late to the party. I think that alternative colors should have defaulted to not being different. Or, it should be optional to turn on. As the release happens now, users will suddenly see alternating colors, and will have to manually configure the colors to get rid of them.

I found my way here while trying to figure out how to disable alternating colors.  I don't see how to get `bemenu` to list the default `<color>` values, so it's a little difficult to revert to the previous behavior. Did I miss that in the options, or is the source code the best location for that information?

Cloudef:
The colors are here:
https://github.com/Cloudef/bemenu/blob/master/lib/menu.c#L18-L36
I'll gladly accept PR that does not change the colors by default

zdykstra:
> The colors are here: https://github.com/Cloudef/bemenu/blob/master/lib/menu.c#L18-L36 I'll gladly accept PR that does not change the colors by default

Great, thank you! Just to clarify, are you looking for a change of the internal defaults or a new flag that sets the previously used defaults for the user?

quite:
> I'll gladly accept PR that does not change the colors by default

Done and merged.

kdkasad:
These new options aren't documented in the manual page, only in the built-in help text.

Issue #152, opened 2021-02-01
Adjust cursor width


notramo:
I would like to be able to adjust the cursor width with command line options

benjamb:
The same for height would be ideal, it can look a little odd when the `--line-height` option is set.

Issue #181, opened 2021-06-17
How to change cursor shape?


tmpm697:
I like `dmenu` like cursor shape that default seem to be `l`  and bemenu default to block that I don't like.

so how to change cursor shape or hardcode it, looking at block all the time cause diseases.

tmpm697:
@Cloudef Any chance for this to be fixed?

Cloudef:
Not priority right now, but anyone can submit a patch for it.

tmpm697:
How to change color of cursor?

fancycade:
@tmpm697

For example, this will make the cursor black:

```
bemenu-run --ff='#000000'
```

It also makes the filter font black though too.

I do think the colors should be a bit more documented as it isn't exactly easy to figure out.

Perhaps just a basic example on the README with custom colors.

tmpm697:
@Cloudef : Can you give a hint on this?

PR #268, opened 2022-06-02
Add option to specify cursor width

Barbaross93:
This PR adds an option to specify the cursor width.

Closes https://github.com/Cloudef/bemenu/issues/152 and sort-of  implements https://github.com/Cloudef/bemenu/issues/181.

Only I-beam and block can be figured through combination of `--ch` and `--cw`. Underline cursors probably requires a re-write of the current cursor drawing code =/


Cloudef:
Thanks

PR #273, opened 2022-06-29
Fix missed rendering regressions on Wayland (fixes --grab flag)

joanbm:
This PR contains two closely related regression fixes (introduced by commit 5a095705d2 and released in versions 0.6.5+), which affect exclusively the Wayland backend.
Both are due to some missed edge cases when managing the `dirty` flag that was introduced in that commit, and for the most part just cause bemenu not to re-render after the items are loaded when using the `--grab`, so bemenu stays at "Loading..." until an unrelated event happens (e.g. a key press) that causes an actual re-render.

 The issue fixed by the first commit can be reproduced with:
```sh
{ echo one; sleep 1; echo two; } | BEMENU_BACKEND=wayland bemenu --grab
```
And causes bemenu to pause at "Loading..." almost 100% of the time.

The issue fixed by the second commit can be reproduced with:
```sh
{ echo one; echo two; } | BEMENU_BACKEND=wayland bemenu --grab
```
And causes bemenu to sometimes (~15% on my PC) pause at "Loading...". Reproduces more often when having some background activity.

### DIFF
--- a/lib/menu.c
+++ b/lib/menu.c
@@ -177,10 +177,14 @@ bm_menu_set_filter(struct bm_menu *menu, const char *filter)
 {
     assert(menu);
 
+    if (strcmp(menu->filter ? menu->filter : "", filter ? filter : ""))
+        menu->dirty = true;
+
     free(menu->filter);
     menu->filter_size = (filter ? strlen(filter) : 0);
     menu->filter = (menu->filter_size > 0 ? bm_strdup(filter) : NULL);
     menu->curses_cursor = (menu->filter ? bm_utf8_string_screen_width(menu->filter) : 0);
+    menu->dirty |= (menu->cursor != menu->filter_size);
     menu->cursor = menu->filter_size;
 }
 
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -26,7 +26,7 @@ render(struct bm_menu *menu)
 
     struct window *window;
     wl_list_for_each(window, &wayland->windows, link) {
-        if (menu->dirty && window->render_pending)
+        if (window->render_pending)
             bm_wl_window_render(window, wayland->display, menu);
     }
     wl_display_flush(wayland->display);


client/common/common.c
```diff
@@ -516,6 +516,7 @@ menu_with_options(struct client *client)
         // bm_menu_grab_keyboard(menu, true);
         bm_menu_render(menu);
         bm_menu_set_filter(menu, NULL);
+        menu->dirty = true; // Ensure a redraw on the first call to bm_menu_render after the items are loaded
```
Cloudef, line 519: The dirty is not considered public API. I think you should set this in this function instead https://github.com/Cloudef/bemenu/blob/84bccc02a0c851fd4933bae67b2f815e23e1e6e2/lib/menu.c#L175

joanbm: Done. Note that most `bm_menu_set_*` are missing the dirtiness check, which is what initially led me to try to just set the dirty flag and call it a day, but I agree that the ideal approach going forward should be to set it on all `bm_menu_set_*` functions that cause a change on the UI.

lib/menu.c
```diff
@@ -177,10 +177,14 @@ bm_menu_set_filter(struct bm_menu *menu, const char *filter)
...
     assert(menu);
 
+    if (strcmp(menu->filter ? menu->filter : "", filter ? filter : ""))
+        menu->dirty = true;
+
     free(menu->filter);
     menu->filter_size = (filter ? strlen(filter) : 0);
     menu->filter = (menu->filter_size > 0 ? bm_strdup(filter) : NULL);
     menu->curses_cursor = (menu->filter ? bm_utf8_string_screen_width(menu->filter) : 0);
+    menu->dirty |= (menu->cursor != menu->filter_size);
```
joanbm, line 187: (This is not necessary to fix the issue at hand, but just not to leave the dirtiness check incomplete).

Cloudef:
Thanks merged

Issue #275, opened 2022-07-01
Keypresses register on release (Wayland)


and-nissa:
**Info**

Wayland backend, currently running bemenu v0.6.8             

**Issue**

Between 0.6.7 and 0.6.8, keypresses have changed from registering on initial press (0.6.7) to on release (0.6.8).

I didn't change any configuration between updates, so I'm thinking it's a bug.

CameronNemo:
I am noticing something weird with keyboard input on Wayland too.

joanbm:
I can reproduce the issue and it has the same root cause as #274, in some cases the menu was not being rendered until a further event happened. In this case the key press event did not cause a render when it should have, and the render didn't happen until the next event, most likely the key release.

Should be fixed by #279.

Issue #274, opened 2022-06-30
Wayland rendering issues


stacyharper:
Hey there !

There already was some rendering issues and redraw problems with wayland backend. Some frame was not redrawn when necessary. This commit made things even worse for a reason I dont fully understand yet.

```patch
c7a5812 Fix missed renders on Wayland on temporally close events Joan Bruguera 2022-06-29 00:24
diff --git a/lib/renderers/wayland/wayland.c b/lib/renderers/wayland/wayland.c
index bc3d83d..0bf145e 100644
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -26,7 +26,7 @@ render(struct bm_menu *menu)
 
     struct window *window;
     wl_list_for_each(window, &wayland->windows, link) {
-        if (menu->dirty && window->render_pending)
+        if (window->render_pending)
             bm_wl_window_render(window, wayland->display, menu);
     }
     wl_display_flush(wayland->display);

```

I think we need a deep digging to clean this up. I made a video presenting the current state with 0.6.8. I ~then show in the second part~ show in a first part how bemenu works with this patch:

```diff
diff --git a/lib/renderers/wayland/wayland.c b/lib/renderers/wayland/wayland.c
index 0bf145e..e96b6d6 100644
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -26,7 +26,7 @@ render(struct bm_menu *menu)
 
     struct window *window;
     wl_list_for_each(window, &wayland->windows, link) {
-        if (window->render_pending)
+        if (menu->dirty || window->render_pending)
             bm_wl_window_render(window, wayland->display, menu);
     }
     wl_display_flush(wayland->display);
```

The video is here : https://tube.missbanal.net/w/rbi1mntpcG8CNk28PCNQdb

edit: the video actually show the patched version then the upstream one !

Cloudef:
@joanbm

joanbm:
Hmm, thank you for the video, it makes the issue very clear. I can't get any rendering problem while manually using the mouse&keyboard on a PC. But since I lack a touch device, I tried to send some sudden mouse moves with `swaymsg`, and it looks there are some rendering issues (half the events seem to be skipped). See this video:

https://user-images.githubusercontent.com/519004/177003484-8c71153f-5ec1-46d0-a992-e2821cab9c96.mp4

I will take a deeper look later to see if I can figure what's going on.

[...]

PR #279, opened 2022-07-05
Fix Wayland event loop order to avoid missed renders (fixes regressions on #274, #275)

joanbm:
While #273 fixes the rendering issue on Wayland with the `--grab` flag, it unfortunately also introduced some new regressions (#274, #275).

The cause of both issues is that after #273, the render roop (`render` in wayland.c) looked like this:

1. Render windows if `window->render_pending` is set
2. Wait for events [including a scheduled render, i.e. a frame callback]
3. Schedule render if `menu->dirty` is set
4. Handle events (`return` from `render`) and repeat

The problem with this, is that the `dirty` flag can be set in step (4), but is checked in step (3), so we may go to sleep to wait for more events in (2) without scheduling a render if necessary.

To fix this, this PR changes reorders the render loop to:

1. Schedule render if `menu->dirty` is set
2. Wait for events [including a scheduled render, i.e. a frame callback]
3. Render windows if `window->render_pending` is set
4. Handle events (`return` from `render`) and repeat

So the `menu->dirty` flag is checked at the beginning, right after bemenu has finished handling events and maybe needs to be rendered.

With this, I can no longer reproduce #273, #274 (also tested by @stacyharper) nor #275, and hopefully no related rendering issues on Wayland remain.

[Side note: The reason #274 and #275 worked fine before #273 is because the way the render loop logic worked before, `window->render_pending` was almost always true before waiting for events, even when no renderings were happening. This caused the menu to immediately repaint when `menu->dirty` was later set to true, so things mostly worked fine anyway]

This PR also includes a fix for a very minor issue for pointer events I found while testing: An "enter" event did not cause the highlighted index to be updated, only a "motion" event.

### DIFF
--- a/lib/menu.c
+++ b/lib/menu.c
@@ -1190,7 +1190,7 @@ bm_menu_run_with_pointer(struct bm_menu *menu, struct bm_pointer pointer)
         }
     }
 
-    if (pointer.event_mask & POINTER_EVENT_MOTION) {
+    if (pointer.event_mask & (POINTER_EVENT_ENTER | POINTER_EVENT_MOTION)) {
         menu_point_select(menu, pointer.pos_x, pointer.pos_y, displayed);
     }
 
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -14,22 +14,23 @@
 static int efd;
 
 static void
-render(struct bm_menu *menu)
-{
-    struct wayland *wayland = menu->renderer->internal;
-    wl_display_dispatch_pending(wayland->display);
-
-    if (wl_display_flush(wayland->display) < 0 && errno != EAGAIN) {
-        wayland->input.sym = XKB_KEY_Escape;
-        return;
-    }
-
+render_windows_if_pending(const struct bm_menu *menu, struct wayland *wayland) {
     struct window *window;
     wl_list_for_each(window, &wayland->windows, link) {
         if (window->render_pending)
             bm_wl_window_render(window, wayland->display, menu);
     }
     wl_display_flush(wayland->display);
+}
+
+static void
+wait_for_events(struct wayland *wayland) {
+    wl_display_dispatch_pending(wayland->display);
+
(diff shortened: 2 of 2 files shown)


Cloudef:
Merged, thanks

joanbm:
PS: I've noticed a side effect for keyboard-first users from the second commit, "Fix pointer enter event without further motion": When bemenu first loads, if the pointer happens to be on top of the bemenu surface, the highlighted index will not be the first one but the one under the cursor. It's not a bug per-se, but it's somewhat annoying for the case where you just want to select the first item ASAP (e.g. from a list of most recent items).

So far I haven't found anything better to avoid this other than "ignore pointer enter events for the first X milliseconds", I'll look for a more elegant solution.

stacyharper:
@joanbm While trying to fix another issue I may found a good solution. Could you try this patch ? (https://git.sr.ht/~stacyharper/bemenu/commit/ce6f9f359e3a99fced723747a37d8549e68b07f3)

joanbm:
@stacyharper ef1055eccec850daac8b568726317b4f8f0d4c2a handles the scenario where the mouse moves over bemenu and only causes a single pointer enter event but no motion event. It's not a very "real" (usual) scenario but it can be done by quickly moving the pointer over bemenu and suddenly stopping, here's an sample of me doing it with a regular laptop trackpad:

https://user-images.githubusercontent.com/519004/178125622-c25d3907-0ddb-470d-9e39-a1ac7bb12a4f.mp4

It can also be easily simulated/reproduced with e.g. `swaymsg seat - cursor set 123 321`.

You're right, your change will basically revert ef1055eccec850daac8b568726317b4f8f0d4c2a, by not changing the highlighted items pointer enter again.

I'm OK with reverting ef1055eccec850daac8b568726317b4f8f0d4c2a for the time being, it's not conceptually wrong, but the scenario it handles is very marginal and yet it fails to correctly handle the common case where the pointer is already sitting where the menu appears.

stacyharper:
Yap I also encounter the issue and that's very annoying.

I think we could instead wait for a MOTION event to select entries from
pointer. That shoud be doable easyly iirc

PR #280, opened 2022-07-06
Don't alternate colors by default (let ALTERNATE color be same as ITEM)

quite:


rieje:
This is incomplete, right? The following alternates despite no alternating colors set:

echo "hey\nwhat\nnow" | /usr/bin/bemenu --list 10 --line-height 22 \
  --tf "#00a1e1" \
  --nf "#ffffff" \
  --hf "#00ffff"  \
  --scf "#00000080"

quite:
@rieje it does not for me. bemenu --version shows v0.6.14

PR #281, opened 2022-07-08
Trigger pointer selection on button release instead

stacyharper:
This allow successive bemenu to not register the same PRESSED
successive events.


Cloudef:
Thanks, makes sense

PR #287, opened 2022-08-01
Fix exiting when an unexpected Wayland error occurs.

joanbm:
If an unexpected error was returned from a Wayland API during rendering (e.g. from `wl_display_flush`), the code did set `input.sym = XKB_KEY_Escape`, so that the next call to `poll_key` would return `BM_KEY_ESCAPE` and bemenu would quit.

However, this has been broken since #135, because input.key_pending was not set, so the "fake" `XKB_KEY_Escape` is just ignored, bemenu doesn't quit, but instead, it enters an infinite loop and keeps a CPU core at 100% usage.

The "quick fix" would be to just set `input.key_pending` wherever `input.sym` was set to `XKB_KEY_Escape`. However, to make error handling less error-prone, decouple it from input handling and add an error flag to `(bm_menu_)render` instead.

*Context:* On one of my Sway setups, every few days I find a bemenu process on the background hogging a CPU core. I am not sure how to reproduce the issue, but by attaching gdb, I saw that `wl_display_flush` was setting `errno=EPIPE` yet bemenu was not exiting. It *may* have something to do with launching bemenu like `long_running_process | bemenu --grab`.

(The issue can be synthetically reproduced by changing `wl_display_flush(wayland->display) < 0 && errno != EAGAIN` to `true`.)

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -556,7 +556,10 @@ run_menu(const struct client *client, struct bm_menu *menu, void (*item_cb)(cons
     struct bm_touch touch;
     enum bm_run_result status = BM_RUN_RESULT_RUNNING;
     do {
-        bm_menu_render(menu);
+        if (!bm_menu_render(menu)) {
+            status = BM_RUN_RESULT_CANCEL;
+            break;
+        }
         key = bm_menu_poll_key(menu, &unicode);
         pointer = bm_menu_poll_pointer(menu);
         touch = bm_menu_poll_touch(menu);
--- a/lib/bemenu.h
+++ b/lib/bemenu.h
@@ -933,7 +933,7 @@ BM_PUBLIC struct bm_item** bm_menu_get_filtered_items(const struct bm_menu *menu
  *
  * @param menu bm_menu instance to be rendered.
  */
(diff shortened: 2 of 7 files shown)


client/common/common.c
```diff
@@ -533,11 +533,16 @@ menu_with_options(struct client *client)
     if (client->grab) {
         bm_menu_set_filter(menu, "Loading...");
         // bm_menu_grab_keyboard(menu, true);
-        bm_menu_render(menu);
+        if (!bm_menu_render(menu))
+            goto fail;
```
Cloudef, line 537: Would it be okay to not check bm_menu_render here? It would fail later in the real loop in the end.

joanbm: I don't think I'd be a problem to avoid checking here, as the call isn't in a loop and so it can't cause the 100% CPU issue. And as you say, it'll fail in the next `bm_menu_render` call, and quit (or is it even possible that the Wayland IPC miraculously recovers in the meanwhile?).

The only difference I see is that the bemenu process will stick around for a while reading all the items from stdin. But that's actually what was already happening before (even when the `XKB_KEY_Escape` trick worked). So I've gone ahead and removed this check here.

Cloudef:
Thanks

PR #290, opened 2022-08-12
[Wayland] Anchor top edge of menu to center

### LINKED ISSUE #289: Duplicate prompts when combining center and list options (wayland)
Version: v0.6.10

When launching `bemenu-run` with the `--center` and `--list` options, after typing multiple copies of the prompt appear. This doesn't seem to happen if I use just the `--center` or `--list` options separately.

bemenu-run screenshot showing duplicate promps

The duplicate prompts seem to appear after typing a character that reduces the size of the list. In the `firefox` example I get a duplicate after typing the `r` and then again after typing the `e`, both of which reduce the # of results. It doesn't happen every time the list size is reduced though.

matt-allan:
Anchor only the top edge of the menu to the center when --center mode is used instead of centering the entire menu. This
prevents the menu from bouncing around as the list is shortened. The issue this fixes is described in this comment: https://github.com/Cloudef/bemenu/issues/194#issuecomment-939280688

For this to work we only anchor to the top instead of the top and the bottom and set a top margin equal to half of the display width.
The max height of the menu also needs to be halved to prevent drawing a menu that overflows the bottom edge of the screen.

Also fixes #289.

This is what it looks like on the main branch when you combine the `--center` and `--list 100` options:

master-center-list-100

It's not really centered anymore because the list was so tall it pushed it to the top of the screen.

When the list is filtered bemenu moves down the screen because it gets shorter, causing the absolute center to move:

main-center-list-10

With this patch the top edge of the list is centered, and not the entire list:

fixed-center-list-100

fixed-center-list-10\

Open questions:

- If you aren't using `--list` mode it still works fine, but the 'center' is going to be just slightly further down own the page. We could only apply this fix for list mode to avoid that but it will make things a bit more complex. Any preference there?
- Is there a better way to implement this? I tried setting a `margin_top` property on the window when it's first created but that didn't work since the alignment isn't set yet and we need to know that to determine the margin.
- Should we let you specify a Y position in px or as a percentage instead? I'd personally like to have the menu a bit higher up on the screen rather than dead center.

### DIFF
--- a/lib/renderers/wayland/window.c
+++ b/lib/renderers/wayland/window.c
@@ -211,10 +211,8 @@ get_align_anchor(enum bm_align align)
 {
     uint32_t anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT | ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
 
-    if(align == BM_ALIGN_TOP) {
+    if(align == BM_ALIGN_TOP || align == BM_ALIGN_CENTER) {
         anchor |= ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP;
-    } else if(align == BM_ALIGN_CENTER) {
-        anchor |= ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP | ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM;
     } else {
         anchor |= ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM;
     }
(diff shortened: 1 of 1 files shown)


hilgenberg:
Yes, please let us specify Y position as percentage, that would be awesome! :-)
And I really don't think this fixes #289 because that is a repaint problem and -c works as it should, otherwise!

matt-allan:
@hilgenberg it fixes #289 for me because it prevents the top of the prompt from moving down the screen in the first place. If the top of the prompt doesn't move then it's repainted in the same place every time which avoids the duplicate prompt issue.

I don't think it's a real fix for the underlying problem (which appears to be a race condition), but it prevents that bug from being 'activated' if that makes sense.

matt-allan:
I'm not using bemenu at the moment and won't be able to work on this any time soon. If someone wants to work off of this PR feel free!

foxyseta:
@Cloudef any updates on this?

PR #293, opened 2022-09-06
wayland: Damage using buffer coordinates

Nefsen402:
The surface was being damaged based on the buffer size. It's more correct to damage using the buffer coordinates.

This is a drive by commit - I haven't actually built the project but this was incorrect before.

For reference: Doing a once over of the code trying to see if I could find anything that would explain this: https://github.com/swaywm/sway/pull/6844#issuecomment-1206564520 (somebody is complaining that bemenu is lagging but I can't reproduce)


Cloudef:
Looks correct, merging

PR #296, opened 2022-09-07
wayland: Bump version

Nefsen402:
The last PR was almost correct but needed a version bump to wl_compositor.

#sway-devel:
```
15:57:47 <baltazar> Nefsen402: btw, I noticed your pr to bemenu and I wanted to test it out but bemenu no longer works after that commit...
15:58:03 <baltazar> it just dies with wl_display@1: error 1: invalid method 9 (since 3 < 4), object wl_surface@14
15:58:21 <bl4ckb0ne> looks like a bemenu mistake
15:58:24 <Nefsen402> Righto - I was half way worried about that
15:58:40 <Nefsen402> We're using a function newer than needed
15:59:02 <bl4ckb0ne> https://github.com/Cloudef/bemenu/blob/c3a662d0961a696c9739b089bcdfe92cf6a89bfd/lib/renderers/wayland/registry.c#L606
15:59:05 <bl4ckb0ne> should be 4
15:59:25 <bl4ckb0ne> https://github.com/Cloudef/bemenu/blob/c3a662d0961a696c9739b089bcdfe92cf6a89bfd/lib/renderers/wayland/registry.c#L623 can also probably yeet this with wl_output 4 fwiw
15:59:38 <Nefsen402> Don't know what the best solution is, revert and just pass in MAX_INT for damage and remain compatible with the old version
15:59:41 <Nefsen402> or bump the version
16:00:11 <bl4ckb0ne> bump it
16:00:14 <Nefsen402> k
16:01:25 <bl4ckb0ne> hm not sure what sets the version of wl_surface
16:01:36 <bl4ckb0ne> but i guess the wl_compositor
16:01:50 cabal704 has joined
16:01:52 <Nefsen402> Guess I'm gonna build it
16:01:55 <Nefsen402> just to make sure
```

PR #297, opened 2022-09-07
Use wl_output version 4 instead of xdg-output-unstable

Nefsen402:
Unfortanately, CI has some pretty old images where this isn't a thing yet. Let's put this into its own PR and decide later.

Reference: https://github.com/Cloudef/bemenu/pull/296#issuecomment-1239857121


Cloudef:
Could change the workflow to use nix instead and I think the ubuntu warts will go away

bl4ckb0ne:
`wl_output` v4 was introduced in wayland 1.20 [1], which was released over a year ago. I think the CI is using ubuntu 20.04 LTS which is stuck in 1.18. Upgrading to the new LTS (22.04) should fix the issue.

[1]: https://gitlab.freedesktop.org/wayland/wayland/-/commit/9c05e6c475e0f67519d88f285a971b5e51455a1d

PR #304, opened 2022-09-19
Add `--accept-single` flag

RichardKraus:
I think this reasonable behaviour since `--ifne` is probably intended for scripted use where a choice is presented.
A caveat is that the name does not really fit anymore, `--choice` or `--if-choice` would be more fitting but splitting this into two options seems unnecessary and removing the old one probably break some scripts.
Another option would be to implement a real `--choice` where only items present in the input list can even be selected that also early exits if no (or one) items are presented.

### DIFF
--- a/client/bemenu-run.c
+++ b/client/bemenu-run.c
@@ -10,6 +10,7 @@
 static struct client client = {
     .filter_mode = BM_FILTER_MODE_DMENU,
     .title = "bemenu-run",
+    .monitor = -1,
 };
 
 struct paths {
--- a/client/bemenu.c
+++ b/client/bemenu.c
@@ -7,6 +7,7 @@
 static struct client client = {
     .filter_mode = BM_FILTER_MODE_DMENU,
     .title = "bemenu",
+    .monitor = -1,
 };
 
 static void
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -179,7 +179,7 @@ usage(FILE *out, const char *name)
           " -x, --password        hide input.\n"
           " -s, --no-spacing      disable the title spacing on entries.\n"
           " --scrollbar           display scrollbar. (none (default), always, autohide)\n"
-          " --ifne                only display menu if there are items.\n"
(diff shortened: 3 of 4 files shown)


Cloudef:
Even with one item, --ifne gives option to cancel the selection

RichardKraus:
Technically scrollbar does not have an optional argument but a default one.
As `--ifne` is not active by default implementing it like scrollbar would still break existing scripts using `bemenu --ifne` (just as `bemenu --scrollbar` is not valid on its own).

From what ive gathered about getopt, implementing it as an actual `optional_argument` however could be very confusing as `bemenu --scrollbar always` would set `optarg="always"` but `bemenu --ifne choice` would set `optarg=NULL`.
Only setting it as `bemenu  --ifne=choice --scrollbar always ` could be used to pass an option.

If you dont want to change the behaviour of the original flag I think the sane option would be to introduce a new flag `--if-choice`.

RichardKraus:
that sounds reasonable.
does it bother you that the "early exit" options were moved above the `bm_menu_grab_keyboard` call?
I didnt look into it but it seems unnecessary to do if no input is taken anyways.
Im asking since it might have sideeffects that I am unaware of.

[This](https://github.com/Cloudef/bemenu/blob/7da879629132ad6924581d53434cde89124887fb/client/common/common.c#L541) is the part I am referring to:
```c
    // move to here

    bm_menu_set_highlighted_index(menu, client->selected);
    bm_menu_grab_keyboard(menu, true);
    bm_menu_set_filter(menu, client->initial_filter);

    // move this block
    if (client->ifne && !bm_menu_get_items(menu, NULL))
        return BM_RUN_RESULT_CANCEL;
```

Cloudef:
There should be no side effects AFAIK

RichardKraus:
I got a bit sidetracked trying to extend the library to also check if the selection was part of the item list.
Since that turned out more difficult than anticipated I settled for `--accept-single` as it better reflects what its actually doing.

Cloudef:
Thanks looks good

Issue #308, opened 2022-10-07
Wayland, cant pointer click when builded with env variable CFLAGS


stacyharper:
Hey there,

With this new 0.6.11, when builded with `CFLAGS="…" make, mouse click doest select the entry. Touch release continues to works.

I have no idea why yet. I'll remove this build flag for this release on Alpine Linux recipe

Cloudef:
If -g affects the behavior it sounds like there is undefined behavior in the code

stacyharper:
I edited the issue a bit. CFLAGS="" also trigger this. I tried to tweak the alpine recipe to unset CFLAGS and the issue continues. I'm still looking for the cause

stacyharper:
The issue comes from the fact that `0 & 0` is not recognised as a true operator most of the time. (for a reason I dunno)

```c
    if (event->state & WL_POINTER_BUTTON_STATE_PRESSED) { // when releasing, 0 & 0
        bm_pointer.state |= POINTER_STATE_PRESSED; 
    } 
```

This solve the issue:

```diff
diff --git a/lib/renderers/wayland/wayland.c b/lib/renderers/wayland/wayland.c
index f5bb33d..4424b46 100644
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -270,10 +270,9 @@ poll_pointer(const struct bm_menu *menu)
     }
 
     if (event->state & WL_POINTER_BUTTON_STATE_PRESSED) {
-        bm_pointer.state |= POINTER_STATE_PRESSED;
-    }
-    if (event->state & WL_POINTER_BUTTON_STATE_RELEASED) {
-        bm_pointer.state |= POINTER_STATE_RELEASED;
+        bm_pointer.state = POINTER_STATE_PRESSED;
+    } else {
+        bm_pointer.state = POINTER_STATE_RELEASED;
     }
 
     memset(event, 0, sizeof(*event));
```

I added this to debug and it gave a very unexpected outcome:

```diff
+    if (0 & 0)
+        fprintf(stderr, "foo\n");
+    if (1 & 1)
+        fprintf(stderr, "bar\n");
+
     if (event->state & WL_POINTER_BUTTON_STATE_PRESSED) {
         bm_pointer.state = POINTER_STATE_PRESSED;
     } else {

```

```
bar
bar
bar
foo
```

Cloudef:
is bm_pointer.state initialized?

Cloudef:
[...]

PR #310, opened 2022-10-07
fix dangling pointer state on wayland

stacyharper:
With this new version, the pointer click doesnt select the highlighted element.

To fix the issue we initialize correctly the bm_pointer struct.

related: https://github.com/Cloudef/bemenu/issues/308


Cloudef:
Can you also revise the code to not use |= https://github.com/Cloudef/bemenu/issues/308#issuecomment-1271399953 as explained here

stacyharper:
Done !

Cloudef:
Thanks

PR #311, opened 2022-10-10
Add support to disable pointer, touch and keyboard events

### LINKED ISSUE #299: Is it possible to turn off mouse / touch support?
Or maybe compile bemenu without them?

dadav:
Sometimes you don't want the pointer/touch/keyboard to have any influence on the menu (e.g. if you mainly use the keyboard, you don't want the mouse to select an item by accident).

closes #299


Cloudef:
Thanks!

PR #314, opened 2022-10-20
Add basic vim bindings

Nimmidev:
I have added basic vim bindings for me a while ago. I was wondering if it would be something you are interested in having in the "main project" @Cloudef . I am happy to make any adjustments as needed (like removing the logs and adding keybinds to the README).

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -181,7 +181,7 @@ usage(FILE *out, const char *name)
           " -C, --no-cursor       ignore cursor events.\n"
           " -T, --no-touch        ignore touch events.\n"
           " -K, --no-keyboard     ignore keyboard events.\n"
-          " --vim                 use vim bindings.\n"
+          " --binding             use alternative key bindings. Available options: vim\n"
           " --scrollbar           display scrollbar. (none (default), always, autohide)\n"
           " --accept-single       immediately return if there is only one item.\n"
           " --ifne                only display menu if there are items.\n"
@@ -309,7 +309,7 @@ do_getopt(struct client *client, int *argc, char **argv[])
         { "scb",          required_argument, 0, 0x114 },
         { "scf",          required_argument, 0, 0x115 },
         { "bdr",          required_argument, 0, 0x121 },
-        { "vim",          no_argument, 0, 0x128 },
+        { "binding",      required_argument, 0, 0x128 },
 
         { "disco",       no_argument,       0, 0x116 },
         { 0, 0, 0, 0 }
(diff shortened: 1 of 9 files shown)


lib/vim.c
```diff
@@ -0,0 +1,335 @@
...
+#include <string.h>
+#include <ctype.h>
+#include <unistd.h>
+#include <assert.h>
+
+#include "vim.h"
+
+
+#define LOG(x)
+//#define LOG(x) printf(x)
```
Cloudef, line 12: you may want to use fprintf(stderr, x) instead

client/common/common.c
```diff
@@ -181,6 +181,7 @@ usage(FILE *out, const char *name)
           " -C, --no-cursor       ignore cursor events.\n"
           " -T, --no-touch        ignore touch events.\n"
           " -K, --no-keyboard     ignore keyboard events.\n"
+          " --vim                 use vim bindings.\n"
```
Cloudef, line 184: I think we should use --binding option with vim as argument instead. Some people may want to add their own bindings as well and there's already quite a lot of these options.

lib/vim.c
Nimmidev: i removed the logging in 9aee56c3. I just used it for testing in the beginning.

client/common/common.c
Nimmidev: Good call. I changed it in 6e1f83dc. I hope the implementation is satisfactory.

client/common/common.h
```diff
@@ -36,7 +36,7 @@ struct client {
     bool force_fork, fork;
     bool no_exec;
     bool password;
-    bool use_vim_bindings;
+    char *key_binding;
```
Cloudef, line 39: Instead of char* can you make this enum?

Cloudef:
Other than that it looks good. I think with this framework, even the original bindings could be refactored to own file. And perhaps even possibility for a implementation that implements customizable keybindings from a file.

client/common/common.h
Nimmidev: Implemented in 51be5c55

Nimmidev:
Yeah, that would be a good idea for the long term. Maybe i will look into this if i have more time available. We'll see.

Cloudef:
Looks good thanks

rieje:
Is this tested on Wayland? I'm on the latest version 0.6.14 with this feature and am not able to enter normal mode to test any of the differences offered with the setting. Another user [reported the same problem](https://github.com/Cloudef/bemenu/issues/315#issuecomment-1356244033) (I'm also on Arch and Wayland).

Nimmidev:
@rieje see #315

rieje:
Can [confirm](https://github.com/Cloudef/bemenu/issues/315#issuecomment-1485523651) it's working now.

Could you add <C-c> and <C-g> (and maybe even ESC) to terminate bemenu in normal mode? The first two already work in insert mode but don't in vim mode. It feels strange to have to go back to insert mode just to terminate bemenu.

PR #316, opened 2022-10-28
Make pkg-config configurable

jspricke:
Needed for crossbuilds.


Cloudef:
Looks good, thank you

PR #320, opened 2022-11-20
improve exit code docs

qguv:
The description of exit codes in the manual has some misleading phrasing regarding custom exit codes. I hope this makes it a bit clearer.


Cloudef:
Thank you

PR #322, opened 2022-12-15
Fix --filter

robertgzr:
- wayland: don't chomp dirty state when window is pending
- common: actually parse -F using getopt
- Apply initial filter before evaluating accept-single


Cloudef:
Thank you

PR #325, opened 2022-12-23
Add Match/Page Counter(Addresses #204)

### LINKED ISSUE #204: [Feature Request] Match counter and page counter
Match counter, e.g. `7/95` for 7 current matching results out of 95 total. This can be displayed at the end of the line. Like [this](https://youtu.be/WAbkzPGhYFI?t=127)

Also, a counter for current pages out of total is IMO cleaner than a scrollbar. Usually a scrollbar implies the use of mouse but dmenu is keyboard-driven.

dynamicdotvoid:
Add a match/page counter to the right side of the bar. I could add an option to disable this, but it shouldn't be required due to how minimal it is. Example: If there are 12 items showing on the bar and a total of 220 items in the system, the bar would show "[12/220]"

~~I forgot to sync my fork with the main branch before adding the counter commits, so for some reason, there are merge conflicts with the main branch, that I couldn't figure out how to fix.~~ Fixed this, I'm not very smart. Ready to merge.

Closes #204

### DIFF
--- a/lib/menu.c
+++ b/lib/menu.c
@@ -375,34 +375,6 @@ bm_menu_get_border_radius(struct bm_menu* menu)
 }
 
 uint32_t
-bm_menu_set_total_item_count(struct bm_menu* menu, uint32_t total_item_count)
-{
-    assert(menu);
-    menu->total_item_count = total_item_count;
-}
-
-uint32_t
-bm_menu_get_total_item_count(struct bm_menu* menu)
-{
-    assert(menu);
-    return menu->total_item_count;
-}
-
-uint32_t
-bm_menu_set_filtered_item_count(struct bm_menu* menu, uint32_t filtered_item_count)
-{
-    assert(menu);
-    menu->filtered_item_count = filtered_item_count;
-}
-
-uint32_t
-bm_menu_get_filtered_item_count(struct bm_menu* menu)
-{
-    assert(menu);
-    return menu->filtered_item_count;
-}
-
-uint32_t
 bm_menu_get_height(struct bm_menu *menu)
 {
     assert(menu);
@@ -794,15 +766,13 @@ bm_menu_filter(struct bm_menu *menu)
 
     char addition = 0;
     size_t len = (menu->filter ? strlen(menu->filter) : 0);
-    uint32_t count;
 
     if (!len || !menu->items.items || menu->items.count <= 0) {
         list_free_list(&menu->filtered);
         free(menu->old_filter);
         menu->old_filter = NULL;
 
-        count = menu->total_item_count;
-        bm_menu_set_filtered_item_count(menu, count);
+        menu->filtered.count = menu->items.count;
         return;
     }
 
@@ -810,19 +780,17 @@ bm_menu_filter(struct bm_menu *menu)
         size_t oldLen = strlen(menu->old_filter);
         addition = (oldLen < len && !memcmp(menu->old_filter, menu->filter, oldLen));
     }
-    
     if (menu->old_filter && addition && menu->filtered.count <= 0)
         return;
 
     if (menu->old_filter && !strcmp(menu->filter, menu->old_filter))
         return;
 
+    uint32_t count;
     struct bm_item **filtered = filter_func[menu->filter_mode](menu, addition, &count);
 
     list_set_items_no_copy(&menu->filtered, filtered, count);
     bm_menu_set_highlighted_index(menu, 0);
-
-    bm_menu_set_filtered_item_count(menu, count);
 
     free(menu->old_filter);
     menu->old_filter = bm_strdup(menu->filter);
--- a/lib/menu.c
+++ b/lib/menu.c
@@ -447,6 +447,18 @@ bm_menu_get_scrollbar(struct bm_menu *menu)
 }
 
 void
+bm_menu_set_counter(struct bm_menu *menu, bool mode)
+{
+    menu->counter = mode;
+}
+
+bool
+bm_menu_get_counter(struct bm_menu *menu)
+{
+    return menu->counter;
+}
+
+void
 bm_menu_set_align(struct bm_menu *menu, enum bm_align align)
 {
     assert(menu);
--- a/lib/menu.c
+++ b/lib/menu.c
@@ -389,6 +389,20 @@ bm_menu_get_total_item_count(struct bm_menu* menu)
 }
 
 uint32_t
+bm_menu_set_filtered_item_count(struct bm_menu* menu, uint32_t filtered_item_count)
+{
+    assert(menu);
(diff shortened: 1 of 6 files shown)


Cloudef:
Hello, can you rebase your changes as there seems to be conflicts

lib/renderers/cairo_renderer.h
```diff
@@ -455,21 +472,30 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
...
         if (menu->wrap || menu->index + 1 < count) {
-            bm_cairo_color_from_menu_color(menu, BM_COLOR_FILTER_FG, &paint.fg);
-            bm_cairo_color_from_menu_color(menu, BM_COLOR_FILTER_BG, &paint.bg);
             bm_pango_get_text_extents(cairo, &paint, &result, ">");
             paint.pos = (struct pos){ width/cairo->scale - result.x_advance - 2, vpadding + border_size };
             paint.box = (struct box){ 1, 2, vpadding, -vpadding, 0, height };
             bm_cairo_draw_line(cairo, &paint, &result, ">");
         }
+        char counter[1024];
+        sprintf(counter, "[%u/%u]", filtered_item_count, total_item_count);
```
Cloudef, line 484: please use `snprintf(counter, sizeof(counter), ...)` instead

lib/renderers/cairo_renderer.h
```diff
@@ -455,21 +472,30 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
...
+        bm_cairo_color_from_menu_color(menu, BM_COLOR_FILTER_BG, &paint.bg);
         if (menu->wrap || menu->index + 1 < count) {
-            bm_cairo_color_from_menu_color(menu, BM_COLOR_FILTER_FG, &paint.fg);
-            bm_cairo_color_from_menu_color(menu, BM_COLOR_FILTER_BG, &paint.bg);
             bm_pango_get_text_extents(cairo, &paint, &result, ">");
             paint.pos = (struct pos){ width/cairo->scale - result.x_advance - 2, vpadding + border_size };
             paint.box = (struct box){ 1, 2, vpadding, -vpadding, 0, height };
             bm_cairo_draw_line(cairo, &paint, &result, ">");
         }
+        char counter[1024];
```
Cloudef, line 483: 64, 128 or something is probably enough.

lib/menu.c
```diff
@@ -360,6 +360,48 @@ bm_menu_get_border_size(struct bm_menu* menu)
...
+}
+
+uint32_t
+bm_menu_get_border_radius(struct bm_menu* menu)
+{
+    assert(menu);
+    return menu->border_radius;
+}
+
+uint32_t
```
Cloudef, line 377: these don't need to be a public API, simply store this information in the menu structs and write / read them directly.

Cloudef: Also you don't need any of these functions or data in the menu struct, as you can get count of filtered items with `menu->filtered.count` and `menu->items.count`

dynamicdotvoid: I was thinking that I was a bit overcomplicating it for the sake of standardization, I'll look into fixing the code.

dynamicdotvoid: `menu->items.count` works great, but `menu->filtered.count` shows **0** when nothing is typed in. While **0** makes sense logically here as nothing is filtered yet, the requested feature showed "totalitemcount/totalitemcount" if nothing was typed in. I'll most likely have to add some things in the `bm_menu_filter` function(I believe that's the name) to fix this, which I hope is ok with you.

dynamicdotvoid: Just committed a fix for both the issue with 0 and the over complication, I'll look into re-basing it.

Cloudef: I think you can simply use `menu->items.count` if `menu->filter` is null

dynamicdotvoid:
Ready to merge, but the Ubuntu test doesn't appreciate that I used a `char*` in a `const char*` function, and from how long flatpak has taken, I assume the same too. I'll look into fixing the warnings, but besides that it's ready to merge.

dynamicdotvoid:
All checks passed, all unnecessary code removed, re-based, and ready to merge.

lib/menu.c
```diff
@@ -771,14 +771,15 @@ bm_menu_filter(struct bm_menu *menu)
         list_free_list(&menu->filtered);
         free(menu->old_filter);
         menu->old_filter = NULL;
+
+        menu->filtered.count = menu->items.count;
```
Cloudef, line 775: Please don't do this as it will cause inconsistent state. Instead in the `bm_cairo_paint` function assign to `filtered_item_count` like this: `(menu->filter ? menu->filtered.count : total_item_count)`

dynamicdotvoid: Ok, working on it. Thought it wouldn't really work out in the end but it fixed the "bug".

dynamicdotvoid:
Fixed the inconsistent state, if there is no other reviews then it should be ready. Apologies for the many problems my code brought up.

Cloudef:
Thanks, one more final thing. Can you expose the counter through CLI switch. You can look `--scrollbar` for example. I think some people may not want it default.

dynamicdotvoid:
Ah, ok, I mentioned something similar on the original issue. By that do you mean enable it through CLI, or disable it there(does the vast majority of people want it enabled or disabled, or would one or the other be more intuitive)?

Cloudef:
Based on experience before, it's better to have it disabled it by default. Usually there always appears a new issue when something is changed by default.

dynamicdotvoid:
On a side note, as I mentioned in the issue, do you think the counter should read "[matched/total]"(current), "matched/total", or something similar?

dynamicdotvoid:
For the counter option, I would most likely have to add a variable in the API to control it? Does that sound right or am I over-complicating things?

Cloudef:
Yes, you can check the code for scrollbar option as for how it is done. There's API for setting scrollbar display setting, and the CLI flag handling code calls that. As for the display formatting, I leave that for you to decide.

dynamicdotvoid:
While doing some testing, the counter seems to disappear when the `-l` option is specified. I'll look into this, but as I modified the code for the ">" in the right of the screen it is most likely a if statement or something that it should be in.

I'm also committing the changes for the option in the CLI, but let me know if I should replace the `bool` with a `enum` in the code(hard to explain, but hopefully you get it).

dynamicdotvoid:
I beg your pardon for all of this trouble, I didn't expect it to last this long.

dynamicdotvoid:
Fixed the `-l` issue, everything seems to be functional currently.

Cloudef:
Thanks, merged!

PR #326, opened 2022-12-29
Add fixed height option(Addresses #270)

### LINKED ISSUE #270: [Feature Request] fixed height
Now that we can spawn `bemenu` in the center of the screen, it would be useful to have an option that allows for spawning it with a fixed height when using `--list`.

Currently, I spawn `bemenu` in the middle of the screen, with 10 lines. As soon as I type something that yields less than 10 results, the whole menu will snap into a new position. This is visually distracting (and doesn't look too good, in my opinion).

stackform:
Add a option to make the menu have a fixed height while in vertical mode using `--fixed-height`. This prevents the menu from shrinking or expanding while filtering through items.

I modified how the filtered item count is set for the counter(getting the filtered item count directly rather than relying on menu->filtered.count) for the sake of efficiency, and it fixed a bug with the counter in vertical mode.

Closes #270

### DIFF
--- a/lib/renderers/cairo_renderer.h
+++ b/lib/renderers/cairo_renderer.h
@@ -297,7 +297,9 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
     uint32_t border_radius = menu->border_radius;
 
     uint32_t total_item_count = menu->items.count;
-    uint32_t filtered_item_count = (menu->filter ? menu->filtered.count : total_item_count);
+    //uint32_t filtered_item_count = (menu->filter ? menu->filtered.count : total_item_count);
+    uint32_t filtered_item_count;
+    bm_menu_get_filtered_items(menu, &filtered_item_count);
 
     cairo_set_source_rgba(cairo->cr, 0, 0, 0, 0);
     cairo_rectangle(cairo->cr, 0, 0, width, height);
@@ -373,11 +375,8 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
         const uint32_t page = (menu->index / lines) * lines;
         
         for (uint32_t l = 0, i = page; l < lines && posy < max_height; ++i, ++l) {
-
-            /* NOTE FOR FUTURE SELF: A BLANK LINE IS SOMETIMES SPAWNED IF NOT USING FIXED HEIGHT && GRAPHICAL ISSUES */
-
-            if (!menu->fixed_height && i > count) {
-                return;
+            if (!menu->fixed_height && i >= count) {
+                continue;
             }
 
             bool highlighted = false;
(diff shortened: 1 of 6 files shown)


dynamicdotvoid:
@Cloudef Any update on this?

lib/renderers/cairo_renderer.h
```diff
@@ -297,7 +297,9 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
     uint32_t border_radius = menu->border_radius;
 
     uint32_t total_item_count = menu->items.count;
-    uint32_t filtered_item_count = (menu->filter ? menu->filtered.count : total_item_count);
+    //uint32_t filtered_item_count = (menu->filter ? menu->filtered.count : total_item_count);
+    uint32_t filtered_item_count;
+    bm_menu_get_filtered_items(menu, &filtered_item_count);
```
Cloudef, line 302: why was this changed?

lib/renderers/cairo_renderer.h
```diff
@@ -382,18 +392,24 @@ bm_cairo_paint(struct cairo *cairo, uint32_t width, uint32_t max_height, const s
...
-            } else if (i % 2 == 1) {
+            } else if (i < count && i % 2 == 1) {
                 bm_cairo_color_from_menu_color(menu, BM_COLOR_ALTERNATE_FG, &paint.fg);
                 bm_cairo_color_from_menu_color(menu, BM_COLOR_ALTERNATE_BG, &paint.bg);
             } else {
                 bm_cairo_color_from_menu_color(menu, BM_COLOR_ITEM_FG, &paint.fg);
                 bm_cairo_color_from_menu_color(menu, BM_COLOR_ITEM_BG, &paint.bg);
             }
+            
+            char *line_str = "error!";
```
Cloudef, line 406: Don't quite understand why this is initialized as "error!", as the if / else below will always either set it to empty or to a formed message.

dynamicdotvoid: I didn't really think of that, I just thought I should have it there, but looking at it it's always going to get initialized.

lib/renderers/cairo_renderer.h
dynamicdotvoid: I'm unsure about if it was a bug with the match counter or some code that I changed, but occasionally, the filtered items wouldn't update properly, so I decided to get the filtered item count from there and it seemed to fix the bug.

dynamicdotvoid: On second thought, I'm pretty sure there was an issue with the ternary statement that incorrectly set the count to 0 occasionally. I believe this was an issue with the counter itself.

dynamicdotvoid:
Should be ready to merge.

stackform:
@Cloudef Any update on this? It's ready, and I've seen discussion about this feature on other issues.

aarobc:
Please merge, this would be great to have.

Cloudef:
Merged thanks!

adastx:
--fixed-height option is only listed in the --help output, seem to be missing from the manpage.

merrittlj:
I completely missed that. You can create a PR about it if you want, as I may not fix this anytime soon.

adastx:
Created it :) #344.

PR #329, opened 2023-01-21
Fix Makefile to support submodule setups

lelgenio:
If you clone the repository as a submodule, `.git` is a symlink, not a directory, so `make` would fail.


Cloudef:
Thanks

PR #332, opened 2023-02-21
Fix: select last entry if pointer/touch above last index

stacyharper:


lib/menu.c
```diff
@@ -677,7 +677,7 @@ bm_menu_set_highlighted_index(struct bm_menu *menu, uint32_t index)
     bm_menu_get_filtered_items(menu, &count);
 
     if (count <= index)
-        return 0;
+        index = count - 1;
```
Cloudef, line 680: This can fail badly if count is 0

Cloudef: Maybe `(count > 0 ? count - 1 : 0)`

stacyharper: Good catch, thanks!

stacyharper:
I also added another minor patch to select the first entry if pointer/touch above the title line. I think it make things more consistent.

Cloudef:
Thanks

PR #334, opened 2023-03-13
ESC have no effect under Wayland while enabled Vim binding

fiathux:
Hi all,
Sorry for my bad English.
I'm runned bemenu under Sway. then I saw ESC can not been captured when I enabled Vim binding, menu will exit directly after ESC pressed.
I have tested key respond in the code. got the unicode is not zero when ESC been presssed, the actual value is ASCII code of ESC (27).
So I try to fix it.
I have tested this patch. it worked fine under AwesomeWM(X11), console(curses) and Sway.


Cloudef:
Thanks

PR #345, opened 2023-05-28
docs: Add missing items to manpage

Inglorion35:
This PR adds the following items to the manpage:

--fixed-height
--accept-single
--ifne
-c, --center
-B, --border
-R, --border-radius
--ch
--cw
--hp
--bdr

ie. all of the missing items as of today.

### DIFF
--- a/man/bemenu.1.scd.in
+++ b/man/bemenu.1.scd.in
@@ -76,12 +76,18 @@ list of executables under PATH and the selected items are executed.
 *--fixed-height*
 	Prevent the display from changing height on filter.
 
+*--accept-single*
+	Immediately return if there is only one item.
+
 *--binding*
 	Use alternative key bindings. Available options: vim
 
 *--fork*
 	Always *fork*(2) before executing the selections. Disabled by default
 	when using the terminal backend.
+
+*--ifne*
+	Only display menu if there are items.
 
 *--no-exec*
 	Print the selected items to standard output instead of executing them.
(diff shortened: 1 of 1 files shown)


Inglorion35:
Also, would you please consider asking contributors to update the docs whenever they add a feature? It doesn't seem like it would be much extra hassle.

Cloudef:
I think the best way would be to auto-generate --help from the man page

Cloudef:
Thank you for the PR, merged!

Inglorion35:
> I think the best way would be to auto-generate --help from the man page

I assume you mean 'auto-generate the man page from `--help`', since it's the man page that keeps going out of date, and which I just had to update. The output of `--help`, which is defined in clients/common.c, seems to be kept up-to-date by contributors - it's what I updated the man page from.

At any rate, that would require a fair amount of work - a LOT of parsing. I haven't seen a ton of activity on this project lately, so is this something that can be implemented quickly?

Because it would _need_ to be implemented quickly. I only found out that the features I needed existed because I happened to glance at the commit history. Most prospective users will read the manpage, not find what they need, and use something else - and the pool of potential contributors will shrink.

An easier solution might be to just check if our scdoc document is up to date before parsing it into the manpage during the build process, and issue a prominent warning if it isn't. Grep the scdoc document for each option in the output of `--help`, and print the ones that aren't found. That way a contributor can be reminded to update it. (Of course, this assumes that `--help` is up to date - you're always going to have to maintain up-to-date documentation _somewhere_, there's no escaping that.)

Again, I hope something can be implemented quickly.

Inglorion35:
(and if THAT's too much work, you could just, y'know, require contributors to update docs)

PR #348, opened 2023-06-15
Add option to exit bemenu using esc in vim binds normal mode

ItsProfessional:
Currently, if I want to close bemenu when using vim bindings, I have to press escape + q
It is more convenient to be able to exit bemenu by pressing esc twice (once to get into normal mode, another time to exit)


Cloudef:
Thanks

PR #350, opened 2023-07-07
Evaluate filter first

robertgzr:
- Apply initial filter before evalutating accept-single
- man: document filter


Cloudef:
Can you update usage in the common.c as well

Cloudef:
Thank you

Issue #335, opened 2023-03-29
`--accept-single` not working on wayland Arch


octvs:
Hey there,
I can't get this flag to work. 
Here are the relevant information about my setup:
```sh
$ echo $BEMENU_OPTS
-i -H 30 -l 5 --tb #458588 --tf #ebdbb2 --fb #282828 --ff #ebdbb2 --nb #282828 --nf #ebdbb2 --hb #3c3836 --hf #458588 --ab #282828 --af #ebdbb2 --sb '#1d2021' --sf #ebdbb2 --fbb '#1d2021' --fbf #ebdbb2 --accept-single --binding vim
```
```sh
$ cat .config/sway/variables | grep \$menu
set $menu j4-dmenu-desktop --no-generic --term="foot" --dmenu="bemenu" | xargs swaymsg exec --
```
```sh
$ hostnamectl
...
Operating System: Arch Linux                      
Kernel: Linux 6.2.8-arch1-1
...
```
```sh
$ pacman -Qs bemenu
local/bemenu-git 0.6.14.r9.g8e2fd06-1
    Dynamic menu library and client program inspired by dmenu
local/bemenu-wayland-git 0.6.14.r9.g8e2fd06-1
    Wayland (wlroots-based compositors) renderer for bemenu
```
```sh
$ sway --version
sway version 1.8.1
```

Kuchteq:
Can confirm the issue is still present.

PR #357, opened 2023-09-26
Add new flag --auto-select: when there is only one entry left, automaticallty select it

gesellkammer:
This PR fixes --accept-single, see #335

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -189,7 +189,8 @@ usage(FILE *out, const char *name)
           " --accept-single       immediately return if there is only one item.\n"
           " --ifne                only display menu if there are items.\n"
           " --fork                always fork. (bemenu-run)\n"
-          " --no-exec             do not execute command. (bemenu-run)\n\n"
+          " --no-exec             do not execute command. (bemenu-run)\n"
+          " --auto-select         when one entry is left, automatically select it\n\n"
 
           "Use BEMENU_BACKEND env variable to force backend:\n"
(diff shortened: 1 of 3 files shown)


Cloudef:
I think there's confusion what accept-single is supposed to do. The accept-single is supposed to return immediately if there's only one item in the list, this seems to change the behavior that accept-single returns immediately if filter out the list enough that there remains only one item.

gesellkammer:
I thought accept single was the pendant to dmenu's "instant" patch or
rofi's auto select. If this is not the case, would you consider adding such
an option?

Cloudef:
Yes that kind of behavior would be a new option

gesellkammer:
Would you accept a patch which implements that functionality under the flag "--accept-single", mirroring rofi's own option?

Cloudef:
That would break the current expected behavior of --accept-single, patch with different flag name is okay.

Cloudef:
Optional argument for --accept-single could work too as well I guess, something like "always"

gesellkammer:
Reverted the behaviour of --accept-single, added a new flag --auto-select

Cloudef:
Can you also document the flag in the man page, looks good otherwise

gesellkammer:
Thanks

Cloudef:
Looks like we hit string length limit:
>error: string literal of length 4104 exceeds maximum length 4095 that ISO C99 compilers are required to support

Cloudef:
Thanks

Issue #342, opened 2023-05-18
[Feature request] Single instance mode


rieje:
Just a simple request to force single instance (prevent multiple instances being launched). Sometimes multiple instances are launched accidentally (via hotkey, so not that uncommon) and it's not obvious this is the case since they overlap. When the newest instance ends, the other instances need to be dealt with and it's rare a user would want to launch multiple instances at once anyway.

Would prefer not to have to run another binary like pgrep.

Earnestly:
A proper solution to this would be using sockets, much like everything else that implements "single instance mode"; but it really makes no sense to put such a limit on `bemenu`, despite the overlapping issue.

I do make use of this aspect which would otherwise require breaking up pipelines into multiple files and stagger execution.

However it's not difficult to create idempotent scripts, without using `pgrep` as you should never really attempt to manage processes (PIDs) that do not belong to your own (children).  Such things are best used in an interactive capacity, much like `htop` and co.

merrittlj:
I will do more research into this, but I can ensure to add an option to disable the single-instance behavior(if it is default behavior).

Earnestly:
For my uses it's not all that important as any changes such as `--sync` would only be aesthetic at the moment. It may have mechanical need for other use cases but I'm not aware of those.
The beauty of a pipeline is that each instance of bemenu is already initialised and waiting for input long before it comes, that and obviating the need for storing intermediate data outside of the pipe buffer.

I do agree with the issue of multi-instances causing trouble, but I don't think single instance is necessary in this case. Instead it would be nice if the initial instances would be refocused instead of just being left totally non-intractable. It is still possible to interact with these windows and refocus them using `xdotool`, e.g. `xdotool search --class bemenu windowfocus`.

In the long term as there is gradually more movement towards wayland I may end up placing more of my interaction in terminals, overloading them for launching commands instead of menus due to greater consistency across platforms.

Earnestly:
I'd suggest changing bemenu to work with multiple instances, so that each menu gets focus when the last used instance closes and not have "single instance mode" at all. It's really not bemenu's fault that people don't write idempotent launchers.

Issue #324, opened 2022-12-23
Add CONTRIBUTING.md


dynamicdotvoid:
Should I(or others) attempt to make an CONTRIBUTING.md file? I think for new contributors it would help them to keep in mind the style of the project, general guidelines to follow, etc.

PR #358, opened 2023-10-03
Single instance mode(Addresses #324)

merrittlj:
Multiple instances of bemenu running at once can freeze/not grab the keyboard once the first instance is closed. This PR kills other bemenu instances when the most recent instance grabs the keyboard, as multiple instances are most likely accidental(need to spam the hotkey or similar, as described by [issue #342](https://github.com/Cloudef/bemenu/issues/342)) and not helped by bemenu's design.

This can be disabled and reverted back to old behavior via the "--multi-instance-mode" flag.

Closes [issue #342](https://github.com/Cloudef/bemenu/issues/342).

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -8,6 +8,8 @@
 #include <unistd.h>
 #include <getopt.h>
 #include <assert.h>
+#include <fcntl.h>
+#include <sys/file.h>
 
 static void
 disco_trap(int sig)
@@ -188,6 +190,7 @@ usage(FILE *out, const char *name)
           " -e, --vim-esc-exits   exit bemenu when pressing escape in normal mode for vim bindings\n"
           " --accept-single       immediately return if there is only one item.\n"
           " --ifne                only display menu if there are items.\n"
+          " --single-instance     force a single menu instance.\n"
           " --fork                always fork. (bemenu-run)\n"
           " --no-exec             do not execute command. (bemenu-run)\n"
           " --auto-select         when one entry is left, automatically select it\n\n"
@@ -281,6 +284,7 @@ do_getopt(struct client *client, int *argc, char **argv[])
         { "accept-single",no_argument,       0, 0x11a },
         { "auto-select",  no_argument,       0, 0x11b },
         { "ifne",         no_argument,       0, 0x117 },
+        { "single-instance",no_argument,     0, 0x11c},
         { "fork",         no_argument,       0, 0x118 },
         { "no-exec",      no_argument,       0, 0x119 },
         { "bottom",       no_argument,       0, 'b' },
@@ -387,6 +391,9 @@ do_getopt(struct client *client, int *argc, char **argv[])
                 break;
             case 0x117:
                 client->ifne = true;
+                break;
+            case 0x11c:
+                client->single_instance = true;
                 break;
             case 0x118:
                 client->force_fork = true;
@@ -598,6 +605,20 @@ menu_with_options(struct client *client)
 enum bm_run_result
 run_menu(const struct client *client, struct bm_menu *menu, void (*item_cb)(const struct client *client, struct bm_item *item))
 {
+    if (client->single_instance) {
+        char *xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+
+        char buffer[1024];
+        char *bemenu_lock_location = (xdg_runtime_dir == NULL ? "/tmp" : xdg_runtime_dir);
+        snprintf(buffer, sizeof(buffer), "%s/bemenu.lock", bemenu_lock_location);
+        
+        int menu_lock = open(buffer, O_CREAT | O_RDWR | O_CLOEXEC, 0666); // Create a read-write lock file(if not already exists), closes on exec. Creates in tmp due to permissions.
+        if (flock(menu_lock, LOCK_EX | LOCK_NB) == -1) {  // If we cannot set the lock/the lock is already present.
+            fprintf(stderr, "bemenu instance is already running");
+            exit(EXIT_FAILURE);
+        }
+    }
+    
     bm_menu_set_highlighted_index(menu, client->selected);
     bm_menu_grab_keyboard(menu, true);
     bm_menu_set_filter(menu, client->initial_filter);
--- a/client/common/common.h
+++ b/client/common/common.h
@@ -33,6 +33,7 @@ struct client {
     bool accept_single;
     bool auto_select;
     bool ifne;
+    bool single_instance;
     bool no_overlap;
     bool no_spacing;
     bool no_cursor;
--- a/man/bemenu.1.scd.in
+++ b/man/bemenu.1.scd.in
@@ -92,6 +92,9 @@ list of executables under PATH and the selected items are executed.
 *--ifne*
 	Only display menu if there are items.
 
+*--single-instance*
+	Force a single menu instance.
+
 *--no-exec*
 	Print the selected items to standard output instead of executing them.
 


merrittlj:
A more elegant solution for forcing a single instance may be present, but I saw that the main impact of multiple instances was in interference during keyboard grabbing(errors "x11: cannot grab keyboard" with multiple instances without this commit).

Earnestly:
I do have a script which runs two `bemenu`s in parallel using both `-f` and `--ifne`. Will this change mean I am forced to break my pipelines up and store intermediate data in files?

merrittlj:
Can you elaborate on the poor use? I'm not the most experienced with some of these tools.

merrittlj:
I am working on implementing a improved solution to this, do not merge this yet and disregard the above details concerning the previous solution.

Cloudef:
1. I think this should be part of common.c (the cli programs) and not the library unless given new API
2. Use flock instead, something like:
```C
    int lock_fd = open("/var/run/bemenu.lock", O_CREAT | O_RDWR | O_CLOEXEC, 0666);
    errno = 0;
    if (flock(lock_fd, LOCK_EX | LOCK_NB) != 0) {
        errx(EXIT_FAILURE, "bemenu instance is already running");
    }
```
then you do not close this fd, it will be closed by OS when the process closes, this should be race free AFAIK

merrittlj:
Got it. I was planning on doing something similar to what you proposed.

merrittlj:
The code is now updated with the updated solution, with help/suggestions from @Earnestly and @Cloudef . It can optionally be disabled/reverted back to the previous behavior with the command line flag "--multi-instance-mode".

client/common/common.c
```diff
@@ -598,6 +607,15 @@ menu_with_options(struct client *client)
 enum bm_run_result
 run_menu(const struct client *client, struct bm_menu *menu, void (*item_cb)(const struct client *client, struct bm_item *item))
 {
+    if (!client->multi_instance_mode) {
+        int menu_lock = open("/tmp/bemenu.lock", O_CREAT | O_RDWR | O_CLOEXEC, 0666); // Create a read-write lock file(if not already exists), closes on exec. Creates in tmp due to permissions.
```
Cloudef, line 611: Try $XDG_RUNTIME_DIR first, if that is not set fallback to /tmp

client/common/common.c
```diff
@@ -598,6 +607,15 @@ menu_with_options(struct client *client)
...
+        errno = 0;
+
+        if (flock(menu_lock, LOCK_EX | LOCK_NB) == -1) {  // If we cannot set the lock/the lock is already present.
+            errx(EXIT_FAILURE, "bemenu instance is already running");
```
Cloudef, line 615: Lets replace this with 
```
fprintf(stderr, "bemenu instance is already running\n");
exit(EXIT_FAILURE);
```
forgot this project does not use err.h and that header does not exist on macos

client/common/common.c
```diff
@@ -188,6 +192,7 @@ usage(FILE *out, const char *name)
           " -e, --vim-esc-exits   exit bemenu when pressing escape in normal mode for vim bindings\n"
           " --accept-single       immediately return if there is only one item.\n"
           " --ifne                only display menu if there are items.\n"
+          " --multi-instance-mode do not force bemenu to run in only a single instance.\n"
```
Cloudef, line 195: I think this flag can be omitted all together. Can't think of use case where this is needed.

client/common/common.c
```diff
@@ -598,6 +606,28 @@ menu_with_options(struct client *client)
...
+        if (xdg_runtime_dir == NULL) {
+            bemenu_lock_location = "/tmp/bemenu.lock";
+            strcpy(buffer, bemenu_lock_location);
+        } else {
+            strcpy(buffer, xdg_runtime_dir);
+            strcat(buffer, "/bemenu.lock");
+        }
+        
+        int menu_lock = open(buffer, O_CREAT | O_RDWR | O_CLOEXEC, 0666); // Create a read-write lock file(if not already exists), closes on exec. Creates in tmp due to permissions.
+        errno = 0;
```
Cloudef, line 623: This isn't needed as we don't check the errno here. Can get rid of errno.h include as well then

client/common/common.c
```diff
@@ -598,6 +605,26 @@ menu_with_options(struct client *client)
...
+        char *xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+        
+        char *bemenu_lock_location;
+        char buffer[128];
+        if (xdg_runtime_dir == NULL) {
+            bemenu_lock_location = "/tmp/bemenu.lock";
+            strcpy(buffer, bemenu_lock_location);
+        } else {
+            strcpy(buffer, xdg_runtime_dir);
+            strcat(buffer, "/bemenu.lock");
```
Cloudef, line 618: Use
```
snprintf(buffer, sizeof(buffer), "%s/bemenu.lock", xdg_runtime_dir);
```
instead

client/common/common.c
```diff
@@ -598,6 +605,26 @@ menu_with_options(struct client *client)
...
 run_menu(const struct client *client, struct bm_menu *menu, void (*item_cb)(const struct client *client, struct bm_item *item))
 {
+    if (!client->multi_instance_mode) {
+        char *xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+        
+        char *bemenu_lock_location;
+        char buffer[128];
+        if (xdg_runtime_dir == NULL) {
+            bemenu_lock_location = "/tmp/bemenu.lock";
+            strcpy(buffer, bemenu_lock_location);
```
Cloudef, line 615: use ```snprintf(buffer, sizeof(buffer), "/tmp/bemenu.lock");```

client/common/common.c
```diff
@@ -598,6 +605,20 @@ menu_with_options(struct client *client)
 enum bm_run_result
 run_menu(const struct client *client, struct bm_menu *menu, void (*item_cb)(const struct client *client, struct bm_item *item))
 {
+    if (!client->multi_instance_mode) {
+        char *xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+
+        char buffer[128];
```
Cloudef, line 611: This can be 1024 or larger

Cloudef:
About the --multi-instance-mode flag, either drop it all together. Or make this new feature under a flag like --single-instance. I don't myself see a need for flag at all but what does @Earnestly think?

merrittlj:
I believe we should have signle instance mode as the default(multiple instances can be confusing to a new user), and have it as toggleable, as while it may not be immediately useful for some, some people may want to take advantage of multiple instances in some way.

In addition, there's no need to remove user control or options, as they don't have any detrimental impact staying in.

Cloudef:
In that case I would just not have flag at all. Otherwise it's opt-in to the old behavior.

merrittlj:
Do you believe we should reenable multiple instance support later on if it is improved in some way(such as correctly focusing the second instance after the first one closes) to where it is more practical?

merrittlj:
I personally support the idea mentioned by @Earnestly in the issue connected to this pull request. We can leave this pull request open for now, but I can move my focus to better support for multiple instances, rather than avoiding it via a forced single-instance mode.

Cloudef:
Maybe change the flag to --single-instance and we can merge this. This way it does not break existing behavior but can still be used. Regarding the last post on the issue, focus controlling may be impossible under wayland, and even then bemenu really won't know the opening order, that would need a daemon.

merrittlj:
Okay, I see. I will make this change when I can. I can experiment with multiple instance support, but with the issues you mentioned, it may not be possible(I am unable to test on platforms such as Wayland, anyways).

merrittlj:
@Cloudef Should be ready to merge, changed to a flag to enable single-instance behavior.

Cloudef:
Looks good, can you also add the option to the man page.

merrittlj:
man page has been updated, should be ready to merge.

Cloudef:
Thanks

PR #360, opened 2023-10-07
Password flag parameters with indicator option(Addresses #221)

### LINKED ISSUE #221: -x option no feedback
The "hide filter text" option (-x, --password) should give feedback to the user in order for the user to
1. Know that what he is typing is recorded
2. Of the number of characters he is typing

Otherwise, the user is typing his password blindly, while the -x option purpose should just be to hide the password.
So a simple way to give feedback would be to print a stars (****) instead of the real input, like a lot of password input fields do.

merrittlj:
Add parameters to the password flag, as well as add an "indicator" option to this flag to replace the input with asterisks, instead of hiding the input or displaying it normally.

Closes #221

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -176,7 +176,7 @@ usage(FILE *out, const char *name)
           " -p, --prompt          defines the prompt text to be displayed.\n"
           " -P, --prefix          text to show before highlighted item.\n"
           " -I, --index           select item at index automatically.\n"
-          " -x, --password        hide input.\n"
+          " -x, --password        display/hide/replace input. (none (default), hide, indicator)\n"
           " -s, --no-spacing      disable the title spacing on entries.\n"
           " -C, --no-cursor       ignore cursor events.\n"
           " -T, --no-touch        ignore touch events.\n"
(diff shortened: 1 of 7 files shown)


client/common/common.c
```diff
@@ -176,7 +176,7 @@ usage(FILE *out, const char *name)
           " -p, --prompt          defines the prompt text to be displayed.\n"
           " -P, --prefix          text to show before highlighted item.\n"
           " -I, --index           select item at index automatically.\n"
-          " -x, --password        hide input.\n"
+          " -x, --password        display/hide/replace input. (none(default), hide(no displayed text), indicator(replace with asterisks)\n"
```
Cloudef, line 179: add space `none (default)`, I don't think the other options need parenthesis description. Also edit the man page

merrittlj: Where can I edit the man page? "man bemenu" or "man bemenu-run" doesn't seem to work for me.

Cloudef: https://github.com/Cloudef/bemenu/blob/master/man/bemenu.1.scd.in

merrittlj: Does the man page have to be built? Is my problem a common issue?

merrittlj: Ah, sorry. I didn't see the scdoc mention in the README.md.

merrittlj:
man page has been updated.

Cloudef:
Thanks

PR #368, opened 2023-11-07
fix: wayland mmap failed on Sway 1.8.1

stacyharper:
Sway 1.8.1 does not trigger some event as surface.enter before we are asked to render the first frame. This cause this first frame to be rendered with wrong attributes. (but that is better than crashing)

This is solved with Sway master, precisely with commit:

> 7d2e4a51063ac90f950cb44f141ab391cbcaff5f
>
> layer-shell: enter output before surface is mapped
>
> This sends fractional-scale-v1 events before the first configure
> event. That way clients have all of the metadata they need to render
> the first frame.

To ensure we don't fails rendering the first frame, even if the compositor does not provide metadata in a correct order, we initialize the window structure with default values.


Cloudef:
Thanks

PR #373, opened 2023-11-19
Split configuration variables to a `config.h` file. (suckless style)

MouadCharradi:
This would allow for a separation between the configuration part and the logic part of bemenu.

Thanks

### DIFF
--- a/client/bemenu-run.c
+++ b/client/bemenu-run.c
@@ -5,13 +5,7 @@
 #include <unistd.h>
 #include <dirent.h>
 #include <assert.h>
-#include "common/common.h"
-
-static struct client client = {
-    .filter_mode = BM_FILTER_MODE_DMENU,
-    .title = "bemenu-run",
-    .monitor = -1,
-};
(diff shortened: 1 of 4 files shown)


MouadCharradi:
I just start reading the bemenu codebase, maybe there are more configuration variables that I haven't seen yet. Maybe we could put those in the config file as well.

Cloudef:
Mostly the defaults in common.c perhaps (for bemenu and bemenu-run executable)

MouadCharradi:
I have move more values to `config.h` from `bemenu.c` and `bemenu-run.c`. I'm not sure if there are any default variables that are set in `common`, it seems to me that it just handles cli opts and changes behavior accordingly, but I might be wrong.

As for the warnings, I will try to find a way of getting rid of them (maybe through a `flto` option in the `Makefile`).
Let me know if I did anything wrong, or if there is anything else I should change or do, or even abandon this whole idea if it doesn't align with the project's vision.

Thanks. Mouad.

PR #375, opened 2023-11-21
add option to start vim bindings in normal mode

pabloariasal:
First of all thank you for bemenu, i love it :)

This PR introduces a new command line flag `--vim-normal-mode`, which changes the initial mode for vim bindings to normal (instead of insert mode, which is the default). This flags makes sense in certain circumstances where there is only a small set of possible options to select from (2-5), and it is just easier to tap j repeatedly instead of fuzzy-finding.

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -188,6 +188,7 @@ usage(FILE *out, const char *name)
           " --scrollbar           display scrollbar. (none (default), always, autohide)\n"
           " --counter             display a matched/total items counter. (none (default), always)\n"
           " -e, --vim-esc-exits   exit bemenu when pressing escape in normal mode for vim bindings\n"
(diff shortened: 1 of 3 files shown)


merrittlj:
Possibly this could be specified as a parameter for the "--binding" flag(ex: "--binding vim normal-mode")? We may want to change the "-e"/"--vim-esc-exits" flag as well to conform to this, but let's see what @Cloudef thinks about this idea before any modifications are made to either flag.

pabloariasal:
while I can see some benefits in combining all the vim-related options into a single `--binding` flag this also makes the CLI less intuitive. In my experience these multitoken flags are hard to document and use properly. Two separate toggle flags is as easy as it gets. Agree to wait for @Cloudef's opinion.

Issue #81, opened 2020-01-15
bemenu isn't visible when another application is fullscreen


AdrienLemaire:
versions:
* OS: Arch linux
* sway 1:1.2-5
* bemenu-git r303.98f7c80-1

**Expected behavior:** Triggering bemenu while another app is fullscreen should display bemenu on top of the full-screen application

**Current behavior:** bemenu is shown **behind** the fullscreen application.

In web dev, it would be a matter of bumping the css z-index property. In wayland/X11, no idea ^^

Note: the manpage mentionned an `-n` parameter (--no-overlap) for wayland, but using it still doesn't show bemenu while another app is fullscreen

Cloudef:
https://github.com/Cloudef/bemenu/blob/master/lib/renderers/wayland/window.c#L327
Does changing the `ZWLR_LAYER_SHELL_V1_LAYER_TOP` to `ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY`work here? If not, this is a sway / wlr protocol issue and you should open a issue there. You can read how the wayland protocol for bemenu window works here https://github.com/Cloudef/bemenu/blob/master/lib/renderers/wayland/wlr-layer-shell-unstable-v1.xml

AdrienLemaire:
```
$ git clone https://github.com/Cloudef/bemenu.git
$ cd bemenu
$ sed -i 's/V1_LAYER_TOP/V1_LAYER_OVERLAY/g' lib/renderers/wayland/window.c
$ mkdir -p build/usr/lib && cd build
$ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=usr -DCMAKE_INSTALL_LIBDIR=usr/lib -DBEMENU_X11_RENDERER=OFF -DBEMENU_WAYLAND_RENDERER=ON ..
$ make install
$ ls | ./usr/bin/bemenu
```
Menu still hidden behind the fullscreen alacritty window.

After reading sway(5), I tried to add `popup_during_fullscreen` to my sway config, set to "smart" then "leave_fullscreen". It didn't have any effect. 
Indeed, I have the same issue with notify_send (mako), so this is not a bemenu issue.

tmccombs:
@AdrienLemaire did you create an issue in sway or wlroots for this?

AdrienLemaire:
@tmccombs not yet, it's kinda low in my priorities atm ^^

travankor:
[...]

PR #376, opened 2023-12-16
wayland: ensure bemenu is shown on top of fullscreen windows

Consolatis:
Before this patch, wayland compositors that render fullscreen windows between the top and overlay layers would show bemenu behind an active fullscreen window. For other compositors the z-rendering order was undefined.

This change ensures that bemenu will always be on top of other usual windows (including fullscreen ones).

Fixes
- #81

### DIFF
--- a/lib/renderers/wayland/window.c
+++ b/lib/renderers/wayland/window.c
@@ -388,7 +388,8 @@ bm_wl_window_create(struct window *window, struct wl_display *display, struct wl
 {
     assert(window);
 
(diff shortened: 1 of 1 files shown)


Cloudef:
There used to be security issue related to this in swaywm but I guess it's now ok?
https://github.com/swaywm/swaylock/issues/107

Consolatis:
> There used to be security issue related to this in swaylock but I guess it's now ok? [swaywm/swaylock#107](https://github.com/swaywm/swaylock/issues/107)

Indeed, the current locking protocol `ext-session-lock-v1` doesn't use the overlay layer and is completely independent from the layershell protocol.

PR #377, opened 2024-01-20
feat: C-b to go left

MithicSpirit:
This brings the keybinds more in line with readline/emacs-style bindings. C-l to
go left is preserved for backwards compatibility. Notably, the curses backend
already had C-b to go left, but did not have C-l; it has not been changed.


Cloudef:
Thanks

PR #379, opened 2024-01-30
implement wayland fractional scaling

stacyharper:


stacyharper:
With my bemenu env variable, with my colors, it looks even more blurry now that before xD I might have missed something? The change is expected to be subtil. Someone can give it a try?

stacyharper:
> With my bemenu env variable, with my colors, it looks even more blurry now that before xD I might have missed something? The change is expected to be subtil. Someone can give it a try?

That feeling is probably cause by the fact that without fractional support, the compositor subtle downscaling add a bit of sharpness. I'm not sure but maybe we can configure Cairo to render text better.

Cloudef:
> perhaps this is off topic here, but even with just scale=2 and one monitor, bemenu has always been blurry for me..

Does forcing scale with BEMENU_SCALE work (try X11 backend too)

stacyharper:
rebased over the change in #380

Cloudef:
It seems wayland handling of BEMENU_SCALE is broken then

Cloudef:
I reverted this commit because it caused hyprland to throw error

PR #356, opened 2023-09-16
Wayland: fix monitor and scaling detection

stacyharper:
The whole management of monitor_name with the Wayland backend seems heavily
broken, and doesn't allow to (1) select the target output with -m, and (2) to
correctly detect the used output and use correct attributes.

When used with two monitors with different scale values, depending on
the output order, you get a blurry menu.

To fix this we attach a listener to wayland surfaces to store which outputs
it belongs to. This rewrite almost completly the "recreate_windows"
function.

This also drop completly the xdg-output-unstable-v1 protocol, cause v4
already add the "name" event to fetch the output name.

I did not pushed this to support v6, with the new
"preferred_buffer_scale" event, cause anyway we need to fetch the
output max_height.

PR #380, opened 2024-01-31
fix: buffer is not divisible by scale X

stacyharper:
To prevent this, we move where we set the buffer scale just after where we attach the new buffer to the surface. It should make sure the buffer scale, and the attached buffer stay consistent on the first commit.

ref: https://github.com/Cloudef/bemenu/pull/356


stacyharper:
Okay then I mark this as ready

Cloudef:
Thanks

PR #382, opened 2024-02-05
Fix mouse press being ignored on x11

Anskrevy:
Simple fix. Haven't tested, but this is included in dmenu and it respects mouse clicks.


Cloudef:
Thanks

PR #383, opened 2024-02-07
Fixed verry fast repeat when pressing shift+Tab

Xgames123:
When shift is pressed the keysym XKB_KEY_ISO_Left_Tab is returned instead of XKB_KEY_Tab.
Normally this isn't noticeable because shift+Tab just copies the selected item to the filter. But when I was changing the keybinds in my patched version I found this bug.


Cloudef:
Thanks

PR #386, opened 2024-02-17
Fix use after free in `wayland.c`

Anskrevy:
Pretty much the title. Not sure if you also want to just free in `fail` to begin with as it is only called after `window` is allocated. Execution is ultimately aborted so maybe just a simple `goto fail` instead of bothering to free.


Cloudef:
Thanks

PR #390, opened 2024-03-05
wayland: bind to zwlr_layer_shell_v1 version 3

mahkoh:
zwlr_layer_shell_v1_destroy is not available in version 2 of the interface.

I tried to fix this in https://github.com/Cloudef/bemenu/pull/249 but mistakenly thought that the request was added in version 2.


Cloudef:
Thanks!

Issue #385, opened 2024-02-12
Can't paste in bemenu-run


jiffygist:
Errors are:
```
Waiting for clipboard paste program 'wl-paste -t text/plain' failed: No child processes
Error: Can't open display: (null)
Waiting for clipboard paste program 'xclip -t text/plain -out' failed: No child processes
```

Works if I remove [this line](https://github.com/Cloudef/bemenu/blob/34259c5da7db6f8ad8742ee3f53b600f22fe27ee/client/bemenu-run.c#L175).

PR #391, opened 2024-03-09
Move SIGCHLD handler

ndren:
SIG_DFL on SIGCHLD prevents proper functioning of pclose, which the new version of the pasting functionality requires.

Closes https://github.com/Cloudef/bemenu/issues/385

Ref: https://stackoverflow.com/questions/54189212/popen-returns-1-unexpectedly

### DIFF
--- a/client/bemenu-run.c
+++ b/client/bemenu-run.c
@@ -1,7 +1,6 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
-#include <signal.h>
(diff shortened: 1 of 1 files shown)


ndren:
Force pushed to clean up the commit history.

Please let me know if this is the right approach, thanks!

ndren:
I have moved it, thanks for the suggestion!

Cloudef:
Thanks

PR #398, opened 2024-04-22
Check for unused CLI arguments

Ordoviz:
I found it surprising that `$ bemenu --fn monospace 99` does not change the font size.


Earnestly:
You probably shouldn't find it surprising that a flag which takes an argument doesn't mean you can supply multiple arguments. The patch is also incorrect as it's not "meta characters".

Do bemenu users really need to be told how to use their shell, provided they even use bemenu within a shell context?

PR #404, opened 2024-06-15
Fix: Transparency on the filter reveals a rectangle behind it that shouldn't be there.

Xgames123:
before
image

after
image


Xgames123:
What why is the build failing. I didn't even touch that part of the code

Cloudef:
Might be a compiler update

Cloudef:
Does non-transparent bemenu still work in wayland and X11?

Xgames123:
I haven't tested x11 but I don't see why not.
Il test tomorrow

Xgames123:
Tested on wayland and x11 on default config and my config with the transparency removed

Cloudef:
Thanks

PR #405, opened 2024-06-16
Clipboard fixes/improvements

vain:
Hi,

this addresses a couple of issues I saw in the clipboard area.

I could maybe live without b2a657d (could be considered a breaking change), but please see the commit message on why I think this is an improvement.

Thanks!


Cloudef:
Thanks, looks good. b2a657d I think it's bug if the original implementation did not match dmenu.

vain:
That was quick, thanks!

Issue #399, opened 2024-04-24
Border woes


Erothas:
![image](https://github.com/Cloudef/bemenu/assets/55455670/f1535a08-fe4d-4015-9603-72707a99f449)

This is the command I used: `j4-dmenu-desktop --dmenu='bemenu --single-instance -i -l 10 -H 25 -M 850 --tb "#191622" --tf "#e534ebff" --fb "#191622" --ff "#94e2d5" --nb "#191622E6" --nf "#ffffffff" --hb "#e534ebff" --hf "#ffffffff" --ab "#191622E6" --af "#ffffffff" --fn FiraSans --bdr "#e534ebff" -B 2 -R 20'`

So, there are two issues that you could probably already spot right away: 

1. The border is being drawn over the background, leading to the background's colour bleeding over/under.
2. The border is very clearly jagged. The thicker the border, the worse it becomes as well.

Cloudef:
Is this better on HEAD?

Erothas:
Sorry for getting back to you late. Truth be told, I was sick and tired of everything on hyprland (including the WM itself) being jank and a buggy mess. So, I uninstalled it all, and installed KDE instead. However, just for you, I installed everything back, and fully updated my system. I then tried out bemenu.. and hyprland instantly crashed... I logged back in, made a second attempt at invoking bemenu. Hyprland crashes again... 

Yeah, I am done with this. Until my life allows for it again (which could be sooner than I think), I don't have time to troubleshoot, trace and run after crash reports anymore.

Cloudef:
Can you run the bemenu on the KDE? Doesn't have to be hyprland.

Erothas:
I am dumb. Yeah, even installing something temporary in NixOS is as simple as `nix-shell -p bemenu`.

So yeah, it's perfect now. All the problems I noticed have been resolved. Even including the silly bubbly animation that used to happen with bemenu's launch is gone. Thank you!

PR #407, opened 2024-06-18
Border fixes (fixes: #399)

Xgames123:
- fixed border with corner radius not clipping bg
- fixed corners being too thick
- fixed border looking weird when height/width is less than corner radius
- renamed bm_cairo_draw_rounded_path to bm_cairo_rounded_path because it was not actually drawing the path and only setting it as the shape in cairo.

I checked if the changes work with --fixed-height, --list, --line-height on x11 and wayland.

Also what is the reason that anti-aliasing is explicitly disabled in the code?

### DIFF
--- a/lib/renderers/cairo_renderer.h
+++ b/lib/renderers/cairo_renderer.h
@@ -262,7 +262,7 @@ bm_cairo_entry_message(char *entry_text, bool highlighted, uint32_t event_feedba
 }
 
 static inline void
-bm_cairo_draw_rounded_path(cairo_t *cr, double x, double y, double width, double height, double radius)
+bm_cairo_rounded_path(cairo_t *cr, double x, double y, double width, double height, double radius)
 {
     double ajusted_radius = MIN(MIN(radius, height*0.5), width*0.5); /* Prevent border from intersecting itself. */
     double degrees = M_PI / 180;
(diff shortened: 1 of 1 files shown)


lib/renderers/cairo_renderer.h
```diff
@@ -260,17 +262,19 @@ bm_cairo_entry_message(char *entry_text, bool highlighted, uint32_t event_feedba
 }
 
 static inline void
-bm_cairo_draw_rounded_path(cairo_t *cr, double x, double y, double width, double height, double radius)
+bm_cairo_rounded_path(cairo_t *cr, double x, double y, double width, double height, double radius)
 {
+    double ajusted_radius = MIN(MIN(radius, height*0.5), width*0.5); /* Prevent border from intersecting itself. */
```
Cloudef, line 267: Is this a typo? Should it be `adjusted_radius`?

Cloudef:
Not sure about the anti-alias, it was introduced in commit <https://github.com/Cloudef/bemenu/commit/877ca82cc18feded7f9f4e8bc9661dbd506ada88>

lib/renderers/cairo_renderer.h
Xgames123: Fixed

Cloudef:
Thanks.

PR #411, opened 2024-07-07
Correctness fixes (memory leaks, compiler warning, out-of-bounds read)

joanbm:
This PR is a small pack with fixes for 4 memory leaks found using Valgrind, a harmless GCC 14 compiler warning, and an out-of-bounds read when parsing the --list command line argument. None of the changes should affect user-visible behaviour.

The "wayland: Fix seat resource release logic" commit is a bit more interesting because it not only leaks memory, but it can also cause event handlers (e.g. `keyboard_handle_key`) to be called multiple times. I think this is harmless in the current bemenu implementation, but it caused issues on a branch I am working on.

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -363,7 +363,7 @@ do_getopt(struct client *client, int *argc, char **argv[])
             {
                 char *ptr;
                 client->lines = strtol(optarg, &ptr, 10);
-                client->lines_mode = (!strcmp(ptr + 1, "up") ? BM_LINES_UP : BM_LINES_DOWN);
+                client->lines_mode = (*ptr && !strcmp(ptr + 1, "up") ? BM_LINES_UP : BM_LINES_DOWN);
                 break;
             }
             case 'c':
--- a/lib/list.c
+++ b/lib/list.c
@@ -64,7 +64,7 @@ list_set_items(struct list *list, const void *items, uint32_t nmemb, list_free_f
     }
 
     void *new_items;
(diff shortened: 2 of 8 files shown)


Cloudef:
Looks good to me. Thank you.

client/common/common.c
```diff
@@ -363,7 +363,7 @@ do_getopt(struct client *client, int *argc, char **argv[])
             {
                 char *ptr;
                 client->lines = strtol(optarg, &ptr, 10);
-                client->lines_mode = (!strcmp(ptr + 1, "up") ? BM_LINES_UP : BM_LINES_DOWN);
+                client->lines_mode = (*ptr && !strcmp(ptr + 1, "up") ? BM_LINES_UP : BM_LINES_DOWN);
```
Earnestly, line 366: Should this not be `ptr` instead of `*ptr` if the purpose is to check if `ptr` is not NULL?

joanbm: The idea is not to check if `ptr` is not NULL (in fact, `strtol` guarantees that `ptr` is not NULL), but rather to check if `strtol` parsed the entire string, i.e. if `optarg` consists of only digits.

If that's the case then `ptr` will point to the `NUL` terminator, so `*ptr = 0` and after the changes `strcmp(ptr + 1, ...)` will *not* be called.

Previously, in this case it was calling `strcmp(ptr + 1, ...)`, which was an out-of-bounds read because `ptr + 1` is right past the `NUL` terminator.

Issue #426, opened 2024-10-12
Vertical offset of monitor not accounted for in X11


ptwales:
My primary monitor is positioned at `0x420` instead of `0x0`.  The reason why is because my secondary monitor is rotated 90 degrees and at position `1920x0`.  

xrandr example:
```bash
xrandr --output DisplayPort-0 --pos 0x420  --primary
xrandr --output HDMI-A-0      --pos 1920x0 --rotate left
```

When I run `bemenu-run` as is, the menu renders off screen.  When run with `--bottom` it renders 420px up from the bottom of the screen.  With `--center` it renders 420px up from the center of the screen.

Works perfectly in wayland though.  I'd like to include screenshots but I don't know how to fire spectacle with bemenu running since it grabs keyboard input.  I tried `-K` but it only ignores keyboard input, it still grabs it.

PR #427, opened 2024-10-12
Account for vertical offset of monitor in x11 #426

ptwales:
Issue #426 

I can confirm that default, `--center` and `--bottom` work with these changes on my monitor which has a vertical offset of 420px.  I also tested it with the vertical offset removed and it still worked.  I also tested it on my vertical monitor (offset 1920x0) and it still worked.  I don't know if any other edge cases need testing.


Cloudef:
Thank you!

PR #431, opened 2024-10-29
Avoid glitch on first frame

stacyharper:
With the recent wlroots scene-graph api changes, we no longer receive the wl_display@enter output event before the very first frame. Meaning that the first draw max_height is 640. This cause a visual glitch, and a useless loop.

With this change, we use instead the minimum height of the outputs.

### DIFF
--- a/lib/renderers/wayland/wayland.c
+++ b/lib/renderers/wayland/wayland.c
@@ -530,10 +530,19 @@ recreate_windows(const struct bm_menu *menu, struct wayland *wayland)
     window->hmargin_size = menu->hmargin_size;
     window->width_factor = menu->width_factor;
 
+    uint32_t min_height = 0;
+    struct output *output_tmp = NULL;
+    wl_list_for_each(output_tmp, &wayland->outputs, link) {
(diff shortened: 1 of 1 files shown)


Cloudef:
Does this work with other compositors as well?

stacyharper:
I haven't tried, but this is an non-greedy solution. It will works great when having only one wl_output. But can still cause edge case issue, when having multiple, with different dimensions. And when no output are enabled (locked screen), we fallback to 640.

It is not a perfect fix.. We would need a way to know which wl_output we will use, before committing the first frame. A kind of `get_preferred_output`, similar to the `get_preferred_scale` that we receive early in the flow. I'm still investigating if this is possible somehow. I'll also ask Wayland folks if this has been asked already.

stacyharper:
I findout that this MR is still in this state. This patch still feels like a valid improvement. Even if the solution isn't perfect, it should looks great on most of the user situations. Is it possible to merge this as it?

stacyharper:
Closed in favor of #453.

PR #432, opened 2024-11-20
Feat: fuzzy matching flag -z

### LINKED ISSUE #367: [Feature request] Fuzzy find option
I propose a flag to match items using fuzzy find, similar to this [dmenu patch](https://tools.suckless.org/dmenu/patches/fuzzymatch/).

wintermute-cell:
Adds the option to use fuzzy matching when filtering menu items, along with the necessary flag and usage info.

Port of / strongly inspired by: https://tools.suckless.org/dmenu/patches/fuzzymatch/dmenu-fuzzymatch-4.6.diff

Would close https://github.com/Cloudef/bemenu/issues/367

### DIFF
--- a/client/common/common.c
+++ b/client/common/common.c
@@ -172,6 +172,7 @@ usage(FILE *out, const char *name)
           " -h, --help            display this help and exit.\n"
           " -v, --version         display version.\n"
           " -i, --ignorecase      match items case insensitively.\n"
+          " -z, --fuzzy           enable fuzzy matching.\n"
           " -F, --filter          filter entries for a given string before showing the menu.\n"
           " -w, --wrap            wraps cursor selection.\n"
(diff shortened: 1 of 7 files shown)


m0tholith:
Bumping this, totally works and no issues

bitkeen:
@Cloudef could you review this please?

Cloudef:
Sorry for taking so long. It looks good to me.

wintermute-cell:
thank you!

Cloudef:
Hello, I reverted these changes because static analysis found bugs in the code. The CI probably did not find those because it was running older version of the compiler at the time of PR. Also, I think this feature should be implemented as separate function rather than in the default filter_dmenu_fun as dmenu has mechanism for selectable filters here <https://github.com/Cloudef/bemenu/blob/master/lib/menu.c#L44-L47>. If you send a revised PR, I'll review it again.

PR #441, opened 2025-07-22
Fix: Working --center flag on kwin and smithay-based compositors

### LINKED ISSUE #422: `--center` flag not working on kwin and smithay, `-W` flag not working on smithay
The `--center` flag seems to be ignored by compositors not using wlroots, such as plasma's kwin and compositor using smithay (like [niri](https://github.com/YaLTeR/niri/)). As far as I understand, those compositors fully implement the layer shell protocol, so they should meet the requirements for running bemenu. Is the `--center` flag specific to wlroots?

The `-W` flag instead doesn't work under smithay compositors, while it works as expected in plasma (so it may be an issue on smithay's end, I don't know).

See https://github.com/YaLTeR/niri/issues/575

sophrosyne42:
Partially fixes #422.

Setting the centered anchor to 0 fixes the -c/--center flag for smithay-based compositors, and doesn't appear to break it for others (e.g., Hyprland).

The -W/--width-factor flag is still broken. This commit fixes -W when bemenu is centered, but not on other anchor points.

Tested on niri (Smithay-based), and Hyprland.

### DIFF
--- a/lib/renderers/wayland/window.c
+++ b/lib/renderers/wayland/window.c
@@ -226,7 +226,7 @@ get_align_anchor(enum bm_align align)
     if(align == BM_ALIGN_TOP) {
         anchor |= ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP;
     } else if(align == BM_ALIGN_CENTER) {
-        anchor |= ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP | ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM;
+        anchor = 0;
     } else {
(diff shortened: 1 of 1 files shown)


Cloudef:
Checked and this seems to be correct according to spec

> Requests that the compositor anchor the surface to the specified edges and corners. If two orthogonal edges are specified (e.g. 'top' and 'left'), then the anchor point will be the intersection of the edges (e.g. the top left corner of the output); otherwise the anchor point will be centered on that edge, or in the center if none is specified.

> Anchor is double-buffered, see wl_surface.commit.

PR #456, opened 2026-02-08
Fix width-factor for river >=0.3.13

### LINKED ISSUE #449: Wayland code uses roundtrips incorrectly
This code assumes that a configure event has arrived after a roundtrip:

https://github.com/Cloudef/bemenu/blob/44b82f9f3fb9f04acfaa89471dddc05a20e16883/lib/renderers/wayland/window.c#L447-L452

However, the compositor might delay sending a configure event for various reasons. That causes window->width to be overwritten by the previous value.

Any of the existing code using roundtrips is suspicious.

ptwales:
Commit [df131f9a](https://codeberg.org/river/river-classic/commit/df131f9a9d454ca60fa58e85edda5bd1eff16956) to [river-classic](https://codeberg.org/river/river-classic), broke the `--width-factor -W` feature in bemenu, causing bemenu to always render with full width.  From my testing with similar menu applications and sway, it seems to be strictly a bemenu/river-classic incompatibility.  Based on issue #449, I'm inclined to believe that river simply exposed an existing bug in bemenu.

My PR should be considered a hotfix.   It doesn't actually fix #449 because I'm frankly ignorant what a correct wayland round trip would be.  I can say it doesn't break bemenu's width on sway.

### DIFF
--- a/lib/renderers/wayland/window.c
+++ b/lib/renderers/wayland/window.c
@@ -371,9 +371,10 @@ bm_wl_window_set_width(struct window *window, struct wl_display *display, uint32
 
     window->hmargin_size = margin;
     window->width_factor = factor;
+    window->width = get_window_width(window);
 
     zwlr_layer_surface_v1_set_anchor(window->layer_surface, window->align_anchor);
-    zwlr_layer_surface_v1_set_size(window->layer_surface, get_window_width(window), window->height);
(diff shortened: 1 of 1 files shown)


ptwales:
ubuntu CI error is with installing deps.  It seems to be transient.

```
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

